PR #10, opened 2017-03-14
[cleanup] Remove function `vcast_f_vf`.

fpetrogalli-arm:
The function `vcast_f_vf` was used only in the neon32 helper file.


shibatch:
Those functions are meant for debugging.
They are needed when we compile with MSVC.

fpetrogalli-arm:
How can they be used for debugging? I have searched the whole code base for `vcast_f_vf`, an no code (either in the library or in the tests) invokes it. Am I missing something?

shibatch:
It is handy when we see the content of a vector variable.
It can be used in printf debugging.
If we use gcc, we can just put [0] after a vector register to see the first element of the vector.
However, that is not possible with MSVC, since it does not support that syntax.

fpetrogalli-arm:
I see, thanks for the explanation!

PR #11, opened 2017-03-15
[cleanup] Guard MSVC specific functions.

fpetrogalli-arm:
The `vcastf_f_vf` routine is needed only when debugging with
MSVC. This patch enables the function when compiling with MSVC, for
those architectures that MSVC supports.


shibatch:
This patch looks good now.

PR #12, opened 2017-03-15
This patch improves the performance of xlog_u1 and  other log, exp an…

shibatch:
…d pow functions.

### DIFF
--- a/src/arch/helperavx512f.h
+++ b/src/arch/helperavx512f.h
@@ -43,6 +43,8 @@ typedef __m512i vint2;
 
 //
 
+void Sleef_x86CpuID(int32_t out[4], uint32_t eax, uint32_t ecx);
+
 static int cpuSupportsAVX512F() {
     int32_t reg[4];
     Sleef_x86CpuID(reg, 7, 0);
@@ -207,7 +209,18 @@ static INLINE vopmask visnan_vo_vd(vdouble d) {
 
 static INLINE vint vilogbk_vi_vd(vdouble d) { return vrint_vi_vd(_mm512_getexp_pd(d)); }
 
+// vilogb2k_vi_vd is similar to vilogbk_vi_vd, but the argument has to
+// be a normalized FP value.
 static INLINE vint vilogb2k_vi_vd(vdouble d) { return vrint_vi_vd(_mm512_getexp_pd(d)); }
+
+static INLINE vdouble vgetexp_vd_vd(vdouble d) { return _mm512_getexp_pd(d); }
+static INLINE vfloat vgetexp_vf_vf(vfloat d) { return _mm512_getexp_ps(d); }
+
+static INLINE vdouble vgetmant_vd_vd(vdouble d) { return _mm512_getmant_pd(d, _MM_MANT_NORM_p75_1p5, _MM_MANT_SIGN_nan); }
+static INLINE vfloat vgetmant_vf_vf(vfloat d) { return _mm512_getmant_ps(d, _MM_MANT_NORM_p75_1p5, _MM_MANT_SIGN_nan); }
+
+static INLINE vdouble vfixup_vd_vd_vd_vi2_i(vdouble a, vdouble b, vint2 c, int imm) { return _mm512_fixupimm_pd(a, b, c, imm); }
+static INLINE vfloat vfixup_vf_vf_vf_vi2_i(vfloat a, vfloat b, vint2 c, int imm) { return _mm512_fixupimm_ps(a, b, c, imm); }
 
 static INLINE double vcast_d_vd(vdouble v) {
   double s[VECTLENDP];
--- a/src/arch/helperavx512f.h
+++ b/src/arch/helperavx512f.h
@@ -207,6 +207,8 @@ static INLINE vopmask visnan_vo_vd(vdouble d) {
 
 static INLINE vint vilogbk_vi_vd(vdouble d) { return vrint_vi_vd(_mm512_getexp_pd(d)); }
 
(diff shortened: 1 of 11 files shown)


src/libm/sleefsimddp.c
```diff
@@ -1133,18 +1156,51 @@ static INLINE CONST vdouble2 logk(vdouble d) {
...
+#endif
+
+  vdouble r = vadd_vd_vd_vd(s.x, s.y);
+
+#ifndef ENABLE_AVX512F
+  r = vsel_vd_vo_vd_vd(vispinf_vo_vd(d), vcast_vd_d(INFINITY), r);
+  r = vsel_vd_vo_vd_vd(vor_vo_vo_vo(vlt_vo_vd_vd(d, vcast_vd_d(0)), visnan_vo_vd(d)), vcast_vd_d(NAN), r);
+  r = vsel_vd_vo_vd_vd(veq_vo_vd_vd(d, vcast_vd_d(0)), vcast_vd_d(-INFINITY), r);
+#else
+  r = _mm512_fixupimm_pd(r, d, vcast_vi2_i((4 << (2*4)) | (3 << (4*4)) | (5 << (5*4)) | (2 << (6*4))), 0);
```
fpetrogalli-arm: Please, no target specific code in the core library. Factor it out as suggested for the getexp and the getmant.

src/libm/sleefdp.c
```diff
@@ -1135,14 +1154,39 @@ static INLINE CONST Sleef_double2 logk(double d) {
...
+
+  t = 0.1532076988502701353e+0;
+  t = mla(t, x2, 0.1525629051003428716e+0);
+  t = mla(t, x2, 0.1818605932937785996e+0);
+  t = mla(t, x2, 0.2222214519839380009e+0);
+  t = mla(t, x2, 0.2857142932794299317e+0);
+  t = mla(t, x2, 0.3999999999635251990e+0);
+  t = mla(t, x2, 0.6666666666667333541e+0);
+  
+  Sleef_double2 s = ddadd2_d2_d2_d2(ddmul_d2_d2_d(dd(0.693147180559945286226764, 2.319046813846299558417771e-17), (double)e),
```
fpetrogalli-arm: For consistency, I think we should rename the `Sleef_double2` type to `double2`, as `vdouble2` is used for `vdouble`. I avoided commenting this in all the places where you have been using Sleef_double2.

src/arch/helperavx512f.h
```diff
@@ -207,6 +207,8 @@ static INLINE vopmask visnan_vo_vd(vdouble d) {
 
 static INLINE vint vilogbk_vi_vd(vdouble d) { return vrint_vi_vd(_mm512_getexp_pd(d)); }
 
+static INLINE vint vilogb2k_vi_vd(vdouble d) { return vrint_vi_vd(_mm512_getexp_pd(d)); }
```
fpetrogalli-arm: In this target, this is exactly the same as `vilogbk_vi_vd`. Why adding it if it is already there?

src/libm/sleefsimddp.c
shibatch: Substituting this part with a more generic function seems hard. Please check the specification of _mm512_fixupimm_pd and give me some advice for doing so.

src/arch/helperavx512f.h
shibatch: vilogb2k_vi_vd has a relaxed specification than vilogbk_vi_vd. For AVX512F target, the optimal codes happened to be same.

fpetrogalli-arm: Oh, I see, thanks for explaining. Please add your explanation as a comment.

src/libm/sleefsimddp.c
fpetrogalli-arm: I am not saying you should write a generic function that behaves like the AVX512 intrinsic. I would just rewrite like 1200 as

    r = vfixup_vd_vd_vd(r,d);

and defined `vfixup_vd_vd_vd` in `helperavx512.h` as 

    static vdouble  vfixup_vd_vd_vd(vdouble r, vdouble d) {
      return _mm512_fixupimm_pd(r, d, vcast_vi2_i((4 << (2*4)) | (3 << (4*4)) | (5 << (5*4)) | (2 << (6*4))), 0);
    }

shibatch: The third argument cannot be dropped. It specifies how it substitute the value. So, vfixup_vd_vd_vd has to be exactly same as _mm512_fixupimm_pd. Is that okay?

fpetrogalli-arm: I am not saying to drop any of the arguments. I suggested a `vdouble(vdoube,vdouble)`signature because I think `r` and `d` are the only to variable parameters in this line. If you need to invoke he intrinsic in other parts of the code with any other values for parameter 3 and 4, than I agree you need to add more parameters in the `vfixup` function.

src/libm/sleefdp.c
shibatch: This is the problem discussed in the following page.

http://stackoverflow.com/questions/3995940/casting-one-c-structure-into-another

We cannot expose double2 data type in SLEEF API since it is too common.

Using typedef like below did not work.

typedef double2 Sleef_double2;

Do you have any idea to improve this?

fpetrogalli-arm: Oh, I see what you mean. I was not asking to `typedef double2 Sleef_double2`, I was thinking about removing the `Sleef_double2` type in all the code base and replace it with `double2`, but if you think it is not practical, leave it as it is.

PR #14, opened 2017-03-16
This patch fixes potential problems in ulp 1 versions of trig functions.

shibatch:
ddadd and ddadd2 are functions for double-double addition.
ddadd is simpler and faster than ddadd2, but it requires the first argument to be larger the the second argument.
Through my testing, this condition only holds for the 1st, 5th, 6th and 7th addition in argument reduction.
In the previous source code, all additions are carried out by ddadd, which is potentially dangerous.


fpetrogalli-arm:
Please add the different requirements on the parameters of `ddadd` and `ddadd2` as comments to their definitions - this is useful information!

PR #15, opened 2017-03-20
This patch improves performance for asin and acos.

shibatch:
The previous code calculates asin and acos via atan. The new code calculates them directly using a kernel function for asin.
The performance improvement is 100% for asin_u10, 20% for asin_u35, 90% for acos_u10, and 50% for acos_u35.
This patch also includes changes for trig functions that reduces register pressure and theoretically imrpoves performance, although I cannot see performance increase in the benchmark results.

### DIFF
--- a/src/arch/helperadvsimd.h
+++ b/src/arch/helperadvsimd.h
@@ -338,6 +338,10 @@ static INLINE vdouble vsel_vd_vo_vd_vd(vopmask mask, vdouble x, vdouble y) {
   return vbslq_f64(vreinterpretq_u64_u32(mask), x, y);
 }
 
+static INLINE vdouble vrint_vd_vd(vdouble d) {
+  return vrndnq_f64(d);
+}
+
 /****************************************/
 /* int <--> float conversions           */
 /****************************************/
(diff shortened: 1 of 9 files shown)


shibatch:
So, the ratio of the execution time of the previous code to the new code is 200% for asin_u10, 120% for asin_u35, 190% for acos_u10, and 150% for acos_u35.

fpetrogalli-arm:
On 21 Mar 2017, at 08:09, Naoki Shibata <notifications@github.com> wrote:

> So, the ratio of the execution time of the previous code to the new code is 200% for asin_u10, 120% for asin_u35, 190% for acos_u10, and 150% for acos_u35.
>

This looks very good!

Nitpicking: from now on, please use 2x for 200%, 1.2x for 120%, and so on, otherwise there will always be room for misunderstanding the meaning!

Francesco


> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub, or mute the thread.
>

IMPORTANT NOTICE: The contents of this email and any attachments are confidential and may also be privileged. If you are not the intended recipient, please notify the sender immediately and do not disclose the contents to any other person, use it for any purpose, or store or copy the information in any medium. Thank you.

PR #16, opened 2017-03-21
This patch fixes bugs in rounding functions.

shibatch:
They returned an wrong value if the argument is close to (1 << 23) for sp functions and (1 << 52) for dp functions.
This patch also contains other minor bugfixes.


shibatch:
What do you mean by regression tests?
I'm running tester2 for at least a few hours.

fpetrogalli-arm:
A regression test is a test that makes sure that a bug does not show up again in subsequent patches.

For example, this patch fixes a bug in rounding function that was shown by values like (1 <<23) and (1 << 52). You should add a test in your test suite that compute the rounding functions on those values and that makes sure that the results is as correct.

shibatch:
I appreciate your comment. But, basically you are requesting to do a redundant task. Since the last release, I added 17 x 2 functions, and I have been planning to add all of these functions to tester1. However, as I wrote in the issue, there are actually more things to change. If I add only a part of the new functions to tester1, I need to modify them again. So, I rather chose to make a big change on tester1. You are right, and I should have discussed with you about this, but I don't want to waste my time. When I am busy, I cannot spend time for this project at all. and that depends on the season. Japanese fiscal and academic year starts in April, and since then, I will get busier than now. From June to October, I think I won't have time at all. That's why I was hurrying. I hope you understand my situation.

src/libm-tester/tester.c
```diff
@@ -124,3202 +113,4205 @@ uint64_t d2u(double d) {
...
 
-    double xa[] = { +0.0, -0.0 };
-    double ya[] = { 0.5, 1.5, 2.0, 2.5, 4.0, 100000, 100000.5 };
+      double xa[] = { POSITIVE_INFINITY };
+      double ya[] = { 0.5, 1, 1.5, 2.0, 2.5, 3.0, 100000, 100000.5 };
 
-    boolean success = true;
-    for(i=0;i<sizeof(xa)/sizeof(double) && success;i++) {
-      for(j=0;j<sizeof(ya)/sizeof(double) && success;j++) {
-	if (!isPlusZe
```
fpetrogalli-arm: All the `#define cmpDenorm` macros in this file should be `do {} while(0)` guarded, as explained in http://stackoverflow.com/questions/154136/why-use-apparently-meaningless-do-while-and-if-else-statements-in-c-c-macros

A simple sanity check that avoid introducing bugs that would be quite tricky to debug.

fpetrogalli-arm: Same as before, wrap the macro body into a `do {...} while (0)` loop.

fpetrogalli-arm: Please add the reason for doing this split as a comment.

fpetrogalli-arm: Same, add a comment saying why you do this.

fpetrogalli-arm: Same, comment why.

fpetrogalli-arm: Same, comment why.

fpetrogalli-arm: Same.

fpetrogalli-arm: Same.

fpetrogalli-arm: Same.

fpetrogalli-arm: `do {} while (0)`, here and in all the `checkAccuracy` macro  below.

fpetrogalli-arm: How much of this could be folded into macros? I see tens of loops here that could be written with macros just calling the proper function names... 

Something like the following would save a lot of typing.

```
#ifdef TEST_01
#undef TEST_01
#define TEST_01(name, mpfr_function, child_function) // macro implementation for the first kind of loop
TEST_01("hypot_u35", mpfr_hypot, child_hipot_u35);
TEST_01("hypot_u05", mpfr_hypot, child_hipot_u05);
#undef TEST_01
#endif  // #ifdef TEST_01

#ifdef TEST_02
#undef TEST_02
#define TEST_02(name, mpfr_function, child_function) // macro implementation for another kind of loop
TEST_02("pow", mpfr_pow, child_pow);
// ... 
#undef TEST_02
#endif  // #ifdef TEST_02

```

fpetrogalli-arm: Now that the `showResults` implements early exits, I'd recommend removing all the `goto` statements in the code.

fpetrogalli-arm: `do {} while(0)`, for this and all the `checkAccuracy` macros following.

fpetrogalli-arm: well, this got repetitive enough to become a #define macro with a comment.

```
// comment
#define MEANINGFUL_NAME !(defined(__arm__) && !defined(__aarch64__))
```

How about that?

fpetrogalli-arm:
Hi Naoki - sorry for taking so long in replying, this was a big patch.

I completely understand your position on this change. I finally had some time to review the code and I have to say it is a huge improvement over the previous version.

Unfortunately I just notices that some of the comments are pointing at the line following the one they refer to. I have no idea why this happened - might be my fault, sorry.

src/libm-tester/testerutil.c
```diff
@@ -0,0 +1,230 @@
...
+    return 0;
+  }
+  if (!xisnan(x) && !xisnan(y) && !isinf(x) && !isinf(y)) return sign(x) == sign(y);
+  return 0;
+}
+
+int firstTime = 1;
+mpfr_t fra, frb, frc, frd;
+
+double countULPdp(double d, mpfr_t c) {
```
shibatch: countULP2dp and countULPdp are different. countULP2dp does not report error above DBL_MIN. This is sometimes needed since some functions do not guarantee error bound below DBL_MIN. What do you mean by merging those two functions?

src/libm-tester/testerutil.c
```diff
@@ -0,0 +1,230 @@
...
+
+  mpfr_set_d(frd, d, GMP_RNDN);
+  mpfr_sub(fra, frd, c, GMP_RNDN);
+  mpfr_div(fra, fra, frb, GMP_RNDN);
+  double u = fabs(mpfr_get_d(fra, GMP_RNDN));
+
+  return u + v;
+}
+
+double countULPsp(float d, mpfr_t c) {
```
shibatch: This is same as the difference between countULPdp and countULP2dp. Note that DBL_MIN and FLT_MIN are the smallest NORMALIZED fp values.

src/libm-tester/iutsimd.c
```diff
@@ -190,1364 +190,203 @@ uint32_t f2u(float d) {
...
+      fflush(stdout);						\
+      continue;							\
+    }								\
+  }
+
+#define func_d_d_i(funcStr, funcName) {	\
+    if (startsWith(buf, funcStr " ")) {	\
+      uint64_t u, v;					\
+      sscanf(buf, funcStr " %" PRIx64 " %" PRIx64, &u, &v);		\
+      union {								\
```
shibatch: This looks a simple mistake. I will fix this.
Could you tell me a little bit about your "downstream source?"
I can add support for that if possible.

src/libm-tester/tester.c
shibatch: I am rather considering to replace all the macros with C functions that take the same arguments. A function can be specified with a pointer to the function. Which do you think is better?

shibatch:
No problem. I think we can collaborate better if you tell me a little bit about what happens in your company after I commit a change in the source code at github.

src/libm-tester/tester.c
shibatch: The goal is to make the code readable, and there should be no need to test the tester. The previous code was too long, and as a result, it was hard to understand what is actually tested. So I tried to make it shorter. But, it does not mean that I am happy with the shortest code.

I will appreciate if introducing the google testing framework will improve this situation, but I need to update the current tester anyway.

src/libm-tester/testerutil.c
fpetrogalli-arm: They differ in just these lines - can we pass the DENORMAL_DBL_MIN and DBI_MIN as a function parameter? Same for the double precision ones.
```
mpfr_set_ld(frb, fmaxl(ldexpl(1.0, e-53), DENORMAL_DBL_MIN), GMP_RNDN);
mpfr_set_ld(frb, fmaxl(ldexpl(1.0, e-53), DBL_MIN), GMP_RNDN);
```

shibatch:
What I am thinking is that if I am creating your extra task by committing to github, you would unconsciously hate me doing so. That is not good for the project. I want to make sure that that is not happening.

src/libm-tester/tester.c
fpetrogalli-arm: I see. Writing these tests by merging them into macros would make porting them to other tests system easier. Also, I agree that there is no need to test the tester - I am just thinking that avoiding duplicate code is better. You mentioned replacing macro with C functions and functions pointer - I am fine with that too.

fpetrogalli-arm:
There is no need to worry about giving me extra tasks Naoki - any improvement is welcome. This code is an improvement - I'd be happy to update the downstream version with the new code.

src/libm-tester/testerutil.c
shibatch: I am still thinking if it is better to merge these two functions. This is very delicate and not so trivial. Please let me postpone this decision.

fpetrogalli-arm: If you think it is better to keep them separated, it is fine by me.

src/libm-tester/tester.c
```diff
@@ -124,3202 +113,4205 @@ uint64_t d2u(double d) {
...
+    sprintf(str, funcStr " %" PRIx64 " %" PRIx64 "\n", d2u(arg1), d2u(arg2)); \
+    write(ptoc[1], str, strlen(str));					\
+    if (readln(ctop[0], str, 255) < 1) stop("child " funcStr);		\
+    sscanf(str, "%" PRIx64, &u);					\
+    return u2d(u);							\
   }
-  return 10003;
-}
+
+double child_sin(double x) { child_d_d("sin", x); }
```
shibatch: This would be just a matter of preference, but if the function declaration is included in the macro, the resulting code looks quite different from the standard C syntax.

fpetrogalli-arm: I see what you mean. I am fine with this form, but please wrap the macro body with the `do {...} while(0)` statement as I mentioned in some other comments.

src/libm-tester/tester.c
shibatch: I don't understand what you mean. Is the message printed by fprintf not enough?

fpetrogalli-arm: This is one of the misplaced comments for `#if defined(__arm__) && !defined(__aarch64__)`

fpetrogalli-arm: This is one of the misplaced comments for `#if defined(__arm__) && !defined(__aarch64__)`

fpetrogalli-arm: This is one of the misplaced comments for `#if defined(__arm__) && !defined(__aarch64__)`

fpetrogalli-arm: This is one of the misplaced comments for `#if defined(__arm__) && !defined(__aarch64__)`

fpetrogalli-arm: This is one of the misplaced comments for `#if defined(__arm__) && !defined(__aarch64__)`

fpetrogalli-arm: This is one of the misplaced comments for `#if defined(__arm__) && !defined(__aarch64__)`

fpetrogalli-arm: This is one of the misplaced comments for `#if defined(__arm__) && !defined(__aarch64__)`

shibatch: This is because aarch32 neon is not IEEE754-compliant.

PR #22, opened 2017-04-24
This patch introduces a conditional branch to argument reduction in t…

shibatch:
…rig functions.

When all the elements in the argument vector are less than 15, a simpler and faster reduction is used.
With this patch, the computation speed of AVX2 ulp 1 trig functions is very close to SVML when the arguments are small.
Please see https://www.dropbox.com/s/66pwjhhjl4glfab/i7-6700-170422.xlsx?dl=0 for the performance graphs.

The ratio of the execution time of the previous code to the new code is 1.8x for Sleef_sind4_u10avx2, 1.7x for Sleef_cosd4_u10avx2, 1.5x for Sleef_sincosd4_u10avx2, and 1.8x  for Sleef_tand4_u10avx2, when the argument is in [0, 6.28].

This patch also improves performance and argument domains of xfmod and xfmodf.

### DIFF
--- a/src/arch/helperadvsimd.h
+++ b/src/arch/helperadvsimd.h
@@ -44,11 +44,15 @@ static INLINE int vavailability_i(int name) { return 3; }
 static INLINE void vprefetch_v_p(const void *ptr) { }
 
 static INLINE int vtestallones_i_vo32(vopmask g) {
-  return vget_lane_u32(vpmin_u32(vand_u32(vget_low_u32(g), vget_high_u32(g)), vand_u32(vget_low_u32(g), vget_high_u32(g))), 0);
+  uint32x2_t x0 = vand_u32(vget_low_u32(g), vget_high_u32(g));
+  uint32x2_t x1 = vpmin_u32(x0, x0);
+  return vget_lane_u32(x1, 0);
 }
 
 static INLINE int vtestallones_i_vo64(vopmask g) {
-  return vget_lane_u32(vpmin_u32(vand_u32(vget_low_u32(g), vget_high_u32(g)), vand_u32(vget_low_u32(g), vget_high_u32(g))), 0);
+  uint32x2_t x0 = vand_u32(vget_low_u32(g), vget_high_u32(g));
+  uint32x2_t x1 = vpmin_u32(x0, x0);
+  return vget_lane_u32(x1, 0);
 }
 
 // Vector load / store
--- a/src/arch/helperadvsimd.h
+++ b/src/arch/helperadvsimd.h
@@ -43,6 +43,14 @@ typedef int32x2_t vint;
 static INLINE int vavailability_i(int name) { return 3; }
 static INLINE void vprefetch_v_p(const void *ptr) { }
 
+static INLINE int vtestallones_i_vo32(vopmask g) {
+  return vget_lane_u32(vpmin_u32(vand_u32(vget_low_u32(g), vget_high_u32(g)), vand_u32(vget_low_u32(g), vget_high_u32(g))), 0);
+}
+
+static INLINE int vtestallones_i_vo64(vopmask g) {
+  return vget_lane_u32(vpmin_u32(vand_u32(vget_low_u32(g), vget_high_u32(g)), vand_u32(vget_low_u32(g), vget_high_u32(g))), 0);
+}
+
 // Vector load / store
 static INLINE vdouble vload_vd_p(const double *ptr) { return vld1q_f64(ptr); }
 static INLINE vdouble vloadu_vd_p(const double *ptr) { return vld1q_f64(ptr); }
--- a/src/arch/helpersse2.h
+++ b/src/arch/helpersse2.h
@@ -107,6 +107,9 @@ static INLINE int vavailability_i(int name) {
 
 static INLINE void vprefetch_v_p(const void *ptr) { _mm_prefetch(ptr, _MM_HINT_T0); }
 
+static INLINE int vtestallones_i_vo32(vopmask g) { return _mm_movemask_epi8(g) == 0xFFFF; }
+static INLINE int vtestallones_i_vo64(vopmask g) { return _mm_movemask_epi8(g) == 0xFFFF; }
+
 //
 
 static vint2 vloadu_vi2_p(int32_t *p) { return _mm_loadu_si128((__m128i *)p); }
--- a/src/libm/sleefdp.c
+++ b/src/libm/sleefdp.c
@@ -27,6 +27,9 @@
 #define PI_C 1.2246467864107188502e-16
 #define PI_D 1.2736634327021899816e-24
 
+#define PI_A2 3.141592653589793116
+#define PI_B2 1.2246467991473532072e-16
+
 #define M_2_PI_H 0.63661977236758138243
 #define M_2_PI_L -3.9357353350364971764e-17
 
@@ -743,7 +746,7 @@ EXPORT CONST double xsin(double d) {
   double u, s, t = d;
 
   double dqh = trunck(d * (M_1_PI / (1 << 24))) * (double)(1 << 24);
-  int ql = rintk(d * M_1_PI - dqh);
+  int ql = rintk(mla(d, M_1_PI, -dqh));
 
   d = mla(dqh, -PI_A, d);
   d = mla( ql, -PI_A, d);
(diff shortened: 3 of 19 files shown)


src/arch/helperadvsimd.h
```diff
@@ -43,6 +43,14 @@ typedef int32x2_t vint;
 static INLINE int vavailability_i(int name) { return 3; }
 static INLINE void vprefetch_v_p(const void *ptr) { }
 
+static INLINE int vtestallones_i_vo32(vopmask g) {
+  return vget_lane_u32(vpmin_u32(vand_u32(vget_low_u32(g), vget_high_u32(g)), vand_u32(vget_low_u32(g), vget_high_u32(g))), 0);
```
fpetrogalli-arm: I think this could be done more efficiently by doing

```
 uint32x4_t x0 = vpminq_u32(g);
uint32x2_t x1 = vand_u32(vget_low_u32(x0), vget_high_u32(0));
return vget_lane_u32(x1,0);
```

shibatch: Your code does not compile. vpminq_u32 takes two arguments.

vtestallones_i_vo64 and vtestallones_i_vo32 are basically same as _mm_test_all_ones in SSE4.1, but we can assume each int32_t lane is either 0 or -1.

fpetrogalli-arm: I knew I should have compiled it before sending it to you! :)

Here is another implementation:

```
uint32x2_t x0 = vand_u32(vget_low_u32(x0), vget_high_u32(0));
uint32x2_t x1 = vpmin_u32(x0, x0);
return vget_lane_u32(x1,0);
```
Disclaimer: I haven't tested this one too! But it seems to make sense to me, in the hypothesis that the `uint32_t` lanes are either `0x0` or `0xffffffff`.

src/libm/sleefdp.c
```diff
@@ -777,18 +780,25 @@ EXPORT CONST double xsin(double d) {
...
-  u = mla(dqh, -PI_A, d);
-  s = ddadd_d2_d_d  (u,  ql * -PI_A);
-  s = ddadd2_d2_d2_d(s, dqh * -PI_B);
-  s = ddadd2_d2_d2_d(s,  ql * -PI_B);
-  s = ddadd2_d2_d2_d(s, dqh * -PI_C);
-  s = ddadd2_d2_d2_d(s,  ql * -PI_C);
-  s = ddadd_d2_d2_d(s, (dqh + ql) * -PI_D);
+  int ql;
   
+  if (fabsk(d) < 15) {
```
fpetrogalli-arm: It woudl be good to explain in a comment why 15.

src/libm/sleefdp.c
```diff
@@ -849,20 +859,27 @@ EXPORT CONST double xcos(double d) {
...
-  dqh *= 1 << 24;
-
-  u = mla(dqh, -PI_A*0.5, d);
-  s = ddadd2_d2_d_d (u,  ql * (-PI_A*0.5));
-  s = ddadd2_d2_d2_d(s, dqh * (-PI_B*0.5));
-  s = ddadd2_d2_d2_d(s,  ql * (-PI_B*0.5));
-  s = ddadd2_d2_d2_d(s, dqh * (-PI_C*0.5));
-  s = ddadd2_d2_d2_d(s,  ql * (-PI_C*0.5));
-  s = ddadd_d2_d2_d(s, (dqh + ql) * (-PI_D*0.5));
+  if (d < 15) {
```
fpetrogalli-arm: Why 15? Please add a comment. If it is used in many places, you might want to use a `#define`  and add the comment only in there.

src/libm/sleefdp.c
```diff
@@ -941,17 +958,24 @@ EXPORT CONST Sleef_double2 xsincos(double d) {
...
-  u = mla(dqh, -PI_A*0.5, d);
-  s = ddadd_d2_d_d(u, ql * (-PI_A*0.5));
-  s = ddadd2_d2_d2_d(s, dqh * (-PI_B*0.5));
-  s = ddadd2_d2_d2_d(s, ql * (-PI_B*0.5));
-  s = ddadd2_d2_d2_d(s, dqh * (-PI_C*0.5));
-  s = ddadd2_d2_d2_d(s, ql * (-PI_C*0.5));
-  s = ddadd_d2_d2_d(s, (dqh + ql) * (-PI_D*0.5));
+  int ql;
+  
+  if (fabsk(d) < 15) {
```
fpetrogalli-arm: Same, 15.

src/libm/sleefdp.c
```diff
@@ -1143,18 +1167,25 @@ EXPORT CONST double xtan(double d) {
...
+  int ql;
   
-  u = mla(dqh, -PI_A*0.5, d);
-  s = ddadd_d2_d_d  (u,  ql * (-PI_A*0.5));
-  s = ddadd2_d2_d2_d(s, dqh * (-PI_B*0.5));
-  s = ddadd2_d2_d2_d(s,  ql * (-PI_B*0.5));
-  s = ddadd2_d2_d2_d(s, dqh * (-PI_C*0.5));
-  s = ddadd2_d2_d2_d(s,  ql * (-PI_C*0.5));
-  s = ddadd_d2_d2_d(s, (dqh + ql) * (-PI_D*0.5));
+  if (fabsk(d) < 15) {
```
fpetrogalli-arm: 15

src/libm/sleefdp.c
```diff
@@ -1842,9 +1873,10 @@ EXPORT CONST double xfmod(double x, double y) {
   if (de < DBL_MIN) { nu *= 1ULL << 54; de *= 1ULL << 54; s = 1.0 / (1ULL << 54); }
   Sleef_double2 q, r = dd(nu, 0);
 
-  for(int i=0;i<4;i++) {
+  for(int i=0;i < 20;i++) {
```
fpetrogalli-arm: Why 20?

src/libm/sleefsimdsp.c
```diff
@@ -1647,9 +1647,10 @@ EXPORT CONST vfloat xfmodf(vfloat x, vfloat y) {
 
   vfloat2 q, r = vcast_vf2_vf_vf(nu, vcast_vf_f(0));
 
-  for(int i=0;i<6;i++) {
+  for(int i=0;i<10;i++) {
```
fpetrogalli-arm: Same here, would be good to explain why 10.

src/libm/sleefsp.c
```diff
@@ -1532,9 +1532,10 @@ EXPORT CONST float xfmodf(float x, float y) {
   if (de < FLT_MIN) { nu *= 1ULL << 25; de *= 1ULL << 25; s = 1.0f / (1ULL << 25); }
   Sleef_float2 q, r = df(nu, 0);
 
-  for(int i=0;i<6;i++) {
+  for(int i=0;i<10;i++) {
```
fpetrogalli-arm: Same, explain why 10.

src/libm/sleefsimddp.c
```diff
@@ -332,20 +340,30 @@ EXPORT CONST vdouble xcos(vdouble d) {
...
-  s = ddadd2_vd2_vd_vd(u, vmul_vd_vd_vd(dql, vcast_vd_d(-PI_A*0.5)));
-  s = ddadd2_vd2_vd2_vd(s, vmul_vd_vd_vd(dqh, vcast_vd_d(-PI_B*0.5)));
-  s = ddadd2_vd2_vd2_vd(s, vmul_vd_vd_vd(dql, vcast_vd_d(-PI_B*0.5)));
-  s = ddadd2_vd2_vd2_vd(s, vmul_vd_vd_vd(dqh, vcast_vd_d(-PI_C*0.5)));
-  s = ddadd2_vd2_vd2_vd(s, vmul_vd_vd_vd(dql, vcast_vd_d(-PI_C*0.5)));
-  s = ddadd_vd2_vd2_vd(s, vmul_vd_vd_vd(vadd_vd_vd_vd(dqh, dql), vcast_vd_d(-PI_D*0.5)));
+  vint ql;
+  
+  if (vtestallones_i_vo64(vlt_vo_vd_vd(vabs_vd_vd(d), vcast_vd_d(15)))) {
+    vdouble dql = vrint_vd_vd(vmla_vd_vd_vd_vd(d, vcast_vd_d(M_1_PI), vcast_vd_d(-0.5)));
```
fpetrogalli-arm: I think you should use `const` where possible. Here and in all other places where you defined something that does not get modified. If you agree, apply it just to this patch-set, not the whole codebase!

src/libm/sleefdp.c
```diff
@@ -27,6 +27,9 @@
 #define PI_C 1.2246467864107188502e-16
 #define PI_D 1.2736634327021899816e-24
 
+#define PI_A2 3.141592653589793116
```
fpetrogalli-arm: These coefficients are shared by both scalar and simd implementation. I recommend placing  them in a common (private) header file.

src/libm/sleefdp.c
```diff
@@ -777,18 +780,25 @@ EXPORT CONST double xsin(double d) {
...
-  s = ddadd2_d2_d2_d(s,  ql * -PI_C);
-  s = ddadd_d2_d2_d(s, (dqh + ql) * -PI_D);
+  int ql;
   
+  if (fabsk(d) < 15) {
+    ql = rintk(d * M_1_PI);
+    u = mla(ql, -PI_A2, d);
+    s = ddadd_d2_d_d (u,  ql * -PI_B2);
+  } else {
+    double dqh = trunck(d * (M_1_PI / (1 << 24))) * (double)(1 << 24);
```
fpetrogalli-arm: `const`

src/arch/helpersse2.h
```diff
@@ -107,6 +107,9 @@ static INLINE int vavailability_i(int name) {
 
 static INLINE void vprefetch_v_p(const void *ptr) { _mm_prefetch(ptr, _MM_HINT_T0); }
 
+static INLINE int vtestallones_i_vo32(vopmask g) { return _mm_movemask_epi8(g) == 0xFFFF; }
```
shibatch: Below is the reference page of _mm_movemask_epi8. It collects the most significant bit of each 8-bit element, so I need to check if the returned value is 0xffff to see if the all bits are set.

https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=movemask_&expand=3591

fpetrogalli-arm: My bad, sorry. Thanks for explaining.

PR #25, opened 2017-05-15
Enabling sleef to be referenced with gnu vector abi

shibatch:
Please check the current state.


src/libm/sleefsimddp.c
```diff
@@ -1751,6 +1784,7 @@ EXPORT CONST vdouble2 xmodf(vdouble x) {
 
   return ret;
 }
+#endif
```
fpetrogalli-arm: `//endif ENABLE_GNUABI`

src/libm/sleefsimdsp.c
```diff
@@ -1449,6 +1478,7 @@ EXPORT CONST vfloat2 xmodff(vfloat x) {
 
   return ret;
 }
+#endif
```
fpetrogalli-arm: `//#ifndef ENABLE_GNUABI`

src/script/abspath.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+#!/bin/sh
+if [ `uname` = Darwin ]; then
+    find `cd $1; pwd` -name $2 -print
```
fpetrogalli-arm: This is exactly the same as the one in line 8. Please merge it into the `#else` statement.

src/libm/Makefile
```diff
@@ -26,34 +26,44 @@ endif
...
 
 libm.a : $(OBJ)
 	$(FLOCK) libm.a.lock -c 'ar rcs libm.a $(OBJ)'
 
+libgm.a : $(OBJ2)
+	$(FLOCK) libgm.a.lock -c 'ar rcs libgm.a $(OBJ2)'
+
+OBJ.txt : $(OBJ)
+	rm -rf OBJ.txt
+	for i in `echo $(OBJ)`; do ../script/abspath.sh $(@D) $$i >> OBJ.txt; done
```
fpetrogalli-arm: I think you can replace this line with the following
```
for i in `echo $(OBJ)`; do realpath $$i >> OBJ.txt; done
```

src/libm/Makefile
```diff
@@ -26,34 +26,44 @@ endif
...
-all : libm.a sleef.h
+all : libm.a libgm.a sleef.h
 
 libm.a : $(OBJ)
 	$(FLOCK) libm.a.lock -c 'ar rcs libm.a $(OBJ)'
 
+libgm.a : $(OBJ2)
+	$(FLOCK) libgm.a.lock -c 'ar rcs libgm.a $(OBJ2)'
+
+OBJ.txt : $(OBJ)
```
fpetrogalli-arm: Why are you generating this text file? Same question for OBJ2.txt

src/libm-tester/iutsimdmain.c
```diff
@@ -0,0 +1,57 @@
+//          Copyright Naoki Shibata 2010 - 2017.
```
fpetrogalli-arm: Nice. Can you remove the old `tester.c` now?

src/libm/sleefsimdsp.c
```diff
@@ -412,6 +382,63 @@ EXPORT CONST vfloat xcosf_u1(vfloat d) {
   return u;
 }
 
+#ifndef ENABLE_GNUABI
```
fpetrogalli-arm: Why not change the prototype when enabling GNUABI instead of removing the function?

src/libm/Makefile
shibatch: Please see Makefile under lib directory.
In order to build the DLLs for OSX and Windows, I need a list of absolute paths of the object files.
Please tell me if there is a better way.

src/libm/Makefile
shibatch: realpath is not available on Mac OS X.

src/libm-tester/iutsimdmain.c
shibatch: tester.c is still in use. Do you mean another file?
I think I need all files.

src/script/abspath.sh
shibatch: I wrote in this way in order to make it work on Mac.
On Mac, 'uname -o' does not work. != operator in [ does not work either.

fpetrogalli-arm: Make sense, thanks for the explanation.

src/libm-tester/iutsimdmain.c
fpetrogalli-arm: I thought this was a replacement of `tester.c`. If not, why did you add it?

src/libm/Makefile
fpetrogalli-arm: I see, thanks for the explanation. 

I am not aware of  a better way of doing this other than using`cmake` instead of using hand crafted Makefiles. Keep it as it is, we will think about moving to cmake in the future.

src/script/abspath.sh
fpetrogalli-arm: Thanks for the explanation.

src/libm-tester/iutsimdmain.c
shibatch: Below is what I wrote in the commit log,

   iutsimd.c is now divided into two parts. This is needed to make it
   work with icc. In the previous implementation, the whole program is
   compiled with support for extended ISA (e.g. AVX512F), and this
   caused the IUT not to exit with the correct exit code when that
   ISA is not supported on the host.

fpetrogalli-arm: I see. Thanks for the explanation. It is easy to miss important bits of information when the patches are big and cover multiple aspects. It would be better if you could submit pull request for small incremental changes.

shibatch: I am sorry for that, but it would be too much overhead. I will have smaller amount of time I can spend for this project after June, so I think that would solve this problem.

fpetrogalli-arm:
LGTM. Please merge.
Thank you!

PR #27, opened 2017-05-22
Dispatcher and llvm compatibility

shibatch:
@hfinkel Could you also give me some comments on the llvm compatibility mode?


fpetrogalli-arm:
Hello Naoki, I am not sure I understand this patch. What kind of compatibility are you trying to achieve here? Could you please add a high level description for the way you intend to use this in LLVM, and how you have design it for that? Thank you!

shibatch:
The main goal here is to minimize the effort to incorporate the latest version of SLEEF into llvm. Actually, I don't know what the best way to do that is. In this patch, I just renamed the function names. I think it is better to talk about how I should implement this feature.

shibatch:
For the patch I committed on May 22, the largest change is the addition of dispatcher. llvm compatibility is kind of supplementary. I also tried to add testing of exported symbols by checking the output of nm, but I had to remove it since the output of sort command does not match between systems.

For the patch I committed on May 29, addition of new helper that uses AVX2 for 128 bit computation is the main change. This is convenient if some user does not want to care about the vector width or compatibility and thinks 128-bit is just enough. In that case, the dispatcher functions for 128 bit functions can be called instead of the SSE2 functions. The library automatically selects AVX2 functions when available. This feature is also important when comparing the performance to SVML, which has that feature.

For the patch I committed on June 6, addition of support for 32bit x86 Linux and x64 Cygwin is the main change. These two are still used widely, so it is worth supporting them. There is nothing special in supporting 32bit x86 Linux. For Cygwin, we need to note that calling with 256bit AVX arguments is not supported at ABI level yet. So, I had to disable AVX support for Cygwin. I also changed the output of tester from stderr to stdout. This is because this is more convenient when I run many testers in parallel.

src/libm-tester/tester.c
```diff
@@ -29,7 +29,7 @@
 #include "testerutil.h"
 
 void stop(char *mes) {
-  fprintf(stderr, "%s\n", mes);
+  printf("%s\n", mes);
```
fpetrogalli-arm: Why all these fprintf to printf substitution? I think you should create a separate pull request for this. You patch is big, having to review functional changes among non functional changes just slows down teh whole process. Please revert the fprintf changes for this pull request.

src/libm/sleefsimdsp.c
```diff
@@ -1712,6 +1752,27 @@ EXPORT CONST vfloat xfmodf(vfloat x, vfloat y) {
   return ret;
 }
 
+EXPORT CONST vfloat xsinpif_u05(vfloat d) {
```
fpetrogalli-arm: This and the following 4 functions does not look correct. Why do you have these dummy implementations?

shibatch: These are required by the dispatcher.
Anyway, I am currently implementing these functions and I expect it will finish soon.

src/libm/rename_llvm.h
```diff
@@ -0,0 +1,149 @@
+//          Copyright Naoki Shibata 2010 - 2017.
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE.txt or copy at
+//          http://www.boost.org/LICENSE_1_0.txt)
+
+#define xsin __llvm_sin_f64
```
shibatch: I will once retract this feature, and reimplement it after the discussion on how to design this.

src/libm-tester/tester.c
shibatch: Do you have any problem if I change the output of the tester from stderr to stdout?
For me, this is more convenient when I run many testers in parallel. There are many combinations of OS and compiler now, and how to run testers quickly with different configurations is becoming 
important. If this change is not desirable for your side, I will try to come up with another way.

fpetrogalli-arm: No, no problem with that. But it is a change that should have been introduced via a different pull request.

src/libm/rename_llvm.h
fpetrogalli-arm: Yes - it is better not to introduce code before we have agreed on the interface.

PR #28, opened 2017-05-29
* This patch introduces a conditional branch to each ulp1 trig functi…

shibatch:
…ons and improves the input range and the average computation speed.

* This patch also fixes the bug of round and rint that caused incorrect output of these functions in very rare occations.


fpetrogalli-arm:
Good, ii-ne! :)

PR #31, opened 2017-06-14
This patch adds new helper avx2_128.

shibatch:
It also separates functions that only utilize SSE2 and those utilize SSE4.

### DIFF
--- a/src/arch/helperavx2_128.h
+++ b/src/arch/helperavx2_128.h
@@ -1,0 +1,365 @@
+//          Copyright Naoki Shibata 2010 - 2017.
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE.txt or copy at
+//          http://www.boost.org/LICENSE_1_0.txt)
+
+#if CONFIG == 1
+
+#ifndef __AVX2__
+#error Please specify -mavx2.
+#endif
+
+#else
+#error CONFIG macro invalid or not defined
+#endif
+
+#define ENABLE_DP
+#define LOG2VECTLENDP 1
+#define VECTLENDP (1 << LOG2VECTLENDP)
+#define ENABLE_FMA_DP
+
+#define ENABLE_SP
+#define LOG2VECTLENSP (LOG2VECTLENDP+1)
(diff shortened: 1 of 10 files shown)


fpetrogalli-arm:
Could you please remind me why we need a 128 bit version of the AVX2 helper?

shibatch:
This feature is convenient if some user does not want to care about the vector width or compatibility and thinks 128-bit is just enough. In that case, the 128 bit functions with the dispatcher can be called instead of the SSE2 functions. The library automatically selects AVX2 functions when available.

This feature combined with the dispatcher is also important when comparing the performance to SVML, which has the same combination of the features.

shibatch:
Maybe the most important point is that AVX2 is significantly faster than SSE4 even in 128 bit computation.

fpetrogalli-arm:
I was asking the question because I think that the answer would have been very useful for the commit message! :)

shibatch:
Is anyone other than you reading the message?

fpetrogalli-arm:
I don't know if anyone else is reading the message - but this info should be in the message of the commit, it is a standard development practice to describe the changes in the commit. It keeps the entry barrier lower for anybody willing to contribute.

shibatch:
I feel a bit weird about this. In my opinion, making SLEEF a well-known software is important. If it becomes major, people will start reading the code themselves.

If we will succeed in making SLEEF a well-known software, I expect people would be more interested in the floating point computation algorithms, like the new reduction for single precision trig functions, than the dispatcher or the new helper. This is just my anticipation, but I am quite sure about this from the e-mails and messages that I exchanged with people.

So, in my opinion, if we try to help people understand how SLEEF works, we should describe more about the FP part. You usually ask me to describe more about the other parts, and I assume that that is needed in your job. I don't think people will care about the reason why there is 128bit AVX2 helper.

Anyway, I am planning to describe the FP algorithm in the documentation, so this is not a problem, though.

fpetrogalli-arm:
I agree with you that describing the FP algorithm is way more interesting than the commit message about the AVX2 128-bit!

Please go ahead with this, the only suggestion/request I have  for that is to use markdown for the sources of the documentation, so that it renders nicely directly on github.

PR #32, opened 2017-06-20
This patch adds a dispatcher for x86. It also adds tester for the fun…

shibatch:
…ctions with dispatcher.

5 dummy functions(xsinpif_u05, xtgammf_u1, xlgammf_u1, xerff_u1 and xerfcf_u15) are added since they are referenced from the dispatcher.
I will implement these functions soon.

### DIFF
--- a/src/libm-tester/Makefile
+++ b/src/libm-tester/Makefile
@@ -42,7 +42,7 @@ endif
 
 ifeq ($(ARCH), x86_64)
 
-TARGET_IUT += iutsse2 iutsse4 iutavx
+TARGET_IUT += iutsse2 iutsse4 iutavx iutdsp128 iutdsp256
 TARGET_TESTER += tester2sse2dp tester2sse2sp tester2sse4dp tester2sse4sp tester2avxdp tester2avxsp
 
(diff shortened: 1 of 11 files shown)


fpetrogalli-arm:
Please describe in the commit message what the dispatcher is for. Also, make sure that the code of the macros in the dispatcher is wrapped in the `do { ...} while(0)` construct.

shibatch:
Unlike other macros, each macro in the dispatcher defines two functions and a pointer to a function. So, wrapping with do { ...} while(0) is impossible.

fpetrogalli-arm:
Hi Naoki - apologies for misunderstanding your code. This now looks good to me. Before you submit this, I wonder if there is any way we could test it on Travis-CI. Those are x86 machines, if you can come up with something that tests the dispatcher we could make sure that other commits will not break it. 

Also, make sure you "squash and merge" the content of this branch and that the description of the dispatcher gets added in the commit message and in the C sources of the executable (the latter will be picked up by `doxygen` when we will have it setup).

shibatch:
I have already implemented the testers for dispatchers and they are included in this patch. iutdsp128 and iutdsp256 are built and used while doing make test.

PR #34, opened 2017-06-25
This patch adds single precision versions of error functions and gamma…

shibatch:
… functions.

It also adds cospi functions for both precisions.

This patch also changes the behavior of cosine functions so that it returns 1 when the argumet is too large.


fpetrogalli-arm:
There is nothing target-specific in the code that you have added in the `helper*.h` files. Please merge those in a shared header file (probably `df.h`?).

shibatch:
At this point, no target-specific code is added to the helpers, but I am planning to do optimization similar to the double-precision code. I need to make sure if each of optimizations really improves the performance, so it will take some time. So, please leave the helper code as the current state.

fpetrogalli-arm:
This is OK for me then. Just add a comment to each function saying that you intend to optimize in a later time.

fpetrogalli-arm:
Oh - I just noticed that there is a conflict that needs to be resolved before merging. @Naoki, Please take care of this.

PR #38, opened 2017-07-05
[build] Fix makefile.

fpetrogalli-arm:
The environment variables need to be read with $() in the makefile source.


fpetrogalli-arm:
I got it terribly wrong. :)  Closing.

Issue #39, opened 2017-07-05
Build SLEEF (scalar version) with cmake


fpetrogalli-arm:
Here is the design I discuss with @diaena:

1. Build the shared object file `libsleef.so` out of `sleefsp.c` and `sleefdp.c`. For the moment, we can live with a single `CMakeLists.txt` script in `src/libm`.
2. Provide a `sleef.h` header file which exports the un-mangled names of the functions, as in `xsin` for the double precision SIN function, and adds the `Sleef_*` names mapped in  `rename.h` [1] as function attribute "alias" [2].
3. Configure `.travis.yml` to execute the cmake build of the library as an additional step
4. Start a `build-with-cmake.md` document in the `doc` folder.

When this is done, we can add the `sleefdl.c` and `sleefqp.c` to the scalar library, but I will create a separate issue for that.

@shibatch , @hfinkel: let us know if you have any questions.

Francesco

[1] https://github.com/shibatch/sleef/blob/master/src/libm/rename.h
[2] https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes

shibatch:
I don't understand why you need alias attribute. There will be two names for xsin?

fpetrogalli-arm:
@shibatch yes, this will produce two names. I would prefer to expose `xsin` over `Sleef_sin_u35`, as it would make easier for a user to use it. Unfortunately it seems that the alias mechanism doesn't work on OSX, so for now we will not use this functionality.

Are you happy to expose just `xsin`-type names for now? (The idea is then to use the ABI mangled names for the vector versions). Note that  using the alias mechanism would allow to have as many names as we would like for each function - unfortunately the alias mechanism doesn't seems to be portable across the OS that SLEEF supports.

@diaena : I have changed the steps as follows:

1. Build the shared object file `libsleef.so` out of `sleefsp.c` and `sleefdp.c`. Place CMakeLists.txt` script in `src` and recursively in the folders that are needed to build the library.
2. Provide a `sleef.h` header file which exports the un-mangled names of the functions, as in `xsin` (we can change these names if @shibatch prefers to do so).
3. Configure `.travis.yml` to execute the cmake build of the library as an additional step
4. Start a `build-with-cmake.md` document in the `doc` folder.

shibatch:
[...]

PR #40, opened 2017-07-06
[#39] Initial build of libsleef (just dp and sp) with cmake.

diaena:
This change addresses https://github.com/shibatch/sleef/issues/39 :: requirement 1.


shibatch:
I have not run this, but at least you need -fp-contract=off option.

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+set(LIBSLEEF_SRCS sleefdp.c sleefsp.c)
+
+add_library(sleef SHARED ${LIBSLEEF_SRCS})
```
fpetrogalli-arm: Please define a variable that carries the name of the target. 

If possible, I'd rather have all the variables that carry target (libraries and executables) in the root folder cmake script, each of them with a description of what the target is and where the add_[executable|library] command is located in the project. I think this is better than having targets spread around in the project tree tree.

doc/build-with-cmake.md
```diff
@@ -0,0 +1,8 @@
+Introduction
+============
```
fpetrogalli-arm: Nitpicking - I prefer the ATX-stype headers [1], as they make the header level explicit.

[ 1] http://pandoc.org/MANUAL.html#atx-style-headers

fpetrogalli-arm:
I have changed the base for the merge to `cmake-transition`.

doc/build-with-cmake.md
diaena: Turns out I prefer these too, but I didn't know about them until now. Will change these in the next patch.

doc/build-with-cmake.md
```diff
@@ -0,0 +1,77 @@
...
+setting variables when running cmake. See the list of [options and variables](#build-customization)
+for customizing your build.
+
+5. Now that you have the build files created by cmake, proceed from the top
+of the build directory:
+```
+$ make sleef
+```
+Running this command will create the shared library `libsleef`. On UNIX you
+should see the following files in the `my-sleef-build/lib` directory.
```
fpetrogalli-arm: As for before, these names depend on the target OS. Please remove these details.

.travis.yml
```diff
@@ -6,9 +6,11 @@ group: edge
 before_install:
```
fpetrogalli-arm: Your build is failing because travis has an older version of cmake.

A starting point on hot to upgrade cmake on travis-ci: https://github.com/travis-ci/travis-ci/issues/4631

You might want to use ano

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3 -DNDEBUG")
```
diaena: Ok, but do we not want the highest optimisation flag at all times? Should the default build type be RelWithDebInfo then?

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+set(LIBSLEEF_SRCS sleefdp.c sleefsp.c)
+
+add_library("${TARGET_LIBSLEEF}" SHARED ${LIBSLEEF_SRCS})
+
+# Always compile sleef with -ffp-contract and log at configuration time
+message(STATUS "Using option `${STRICTMATHFLAG}` to compile libsleef")
+target_compile_options(${TARGET_LIBSLEEF} PUBLIC ${STRICTMATHFLAG})
+
+# Shared library versioning
+set_target_properties(${TARGET_LIBSLEEF} PROPERTIES VERSION ${SLEEF_VERSION_MAJOR}.${SLEEF_VERSION_MINOR} SOVERSION ${SLEEF_SOVERSION})
```
diaena: This results in the following library and symbolic links:
libsleef.so.3.0
libsleef.so.3 -> libsleef.so.3.0
libsleef.so -> libsleef.so.3

I believe this is standard shared library versioning: using a version layers that maps to the project version (ie SLEEF 3.0) and another separate layer which is the library version (we currently use the MAJOR project version to be the library version too).

I shall add a comment explaining this.

src/libm/CMakeLists.txt
fpetrogalli-arm: I think that the default build type is `-DCMAKE_BUILD_TYPE=Debug`.
If you want to turn on all the optimization flags, the programmer building the package should use `-DCMAKE_BUILD_FLAG=Release`.

I am not keen in changing the default behavior of `cmake`, but I am happy to do so if it turns out that we cannot build sleef with these two standard flavors.

@Naoki will be able to help out with the compiler flags once we get in trouble. For know leave the standard flags of `cmake`. It might be worth setting travis-ci to test the `Release` build, not the `Debug` one.

CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+You can refer to doc/build-with-cmake.md for instructions on how provide a separate build directory.
+Note: Please remove autogenerated file `CMakeCache.txt` and directory `CMakeFiles` in the current directory.")
+endif()
+
+# Set output directories for the library files
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
+
+# Compiler options; note: this list will build incrementally when introducing system detection
+set(STRICTMATHFLAG -ffp-contract=off)
```
diaena: We could but I think it's better that we keep multiple flag variables to refer to specific types of flags (eg FASTMATHFLAG, SHAREDFLAGS, etc). 
Also, STRICTMATHFLAG may have different values when using different compilers and this value will be determined and (most importantly) cached at detection time.

src/libm/CMakeLists.txt
shibatch: The minor version should be 1 now.

This versioning is also important in packaging for debian. I set up a launchpad page so that compiled binary of sleef can be installed to ubuntu systems with apt-get when 3.0 was released. However, it is not working well since I haven't got used to debian packaging system.

https://code.launchpad.net/~shibatch/+archive/ubuntu/sleef

I think we need a good way to manage files under debian directory.

fpetrogalli-arm:
@Naoki - yes, it includes building on all OSs and with all compilers, to reach the same degree of functionality of the original `Makefile` system.

To facilitate reviews, let's limit the content of this pull request to `gcc` on linux (current travis-ci configuration). We will extent it to other systems in subsequent pull requests.

CMakeLists.txt
fpetrogalli-arm: Thanks for the explanation. Leave the original `FASTMATHFLAGS` name.

src/libm/CMakeLists.txt
fpetrogalli-arm: Oh - this is very good to have. An easy way to install it on a debian-based system. We should work on this on a branch based on `master`, and generate a separate pull request for it. @Naoki, do you mind creating an issue with references to the launchpad page and your requirements?

shibatch:
Although this is not a relevant topic, but ID naoki is used by another person. The ID shibatch may sound weird for you, but it is common in Japan to call a person by his/her sir name, and shibatch is actually a nickname given by my friends when I was an undergraduate student. I decided to use that ID since I started contributing to OSS projects, since absolutely no one seemed using that ID at that time. A weird ID is good, since I can dominate the search results. Try googling "shibatch" with incognito mode.

fpetrogalli-arm:
@shibatch - apologies for the mistake! I have updated the comments.

.travis.yml
```diff
@@ -4,11 +4,19 @@ dist: trusty
 group: edge
 
 before_install:
+    - sudo add-apt-repository ppa:adrozdoff/cmake
     - sudo apt-get -qq update
-    - sudo apt-get install -y libmpfr-dev libgomp1 gcc
-    
+    - sudo apt-get install -y clang cmakelibmpfr-dev libgomp1 gcc
```
fpetrogalli-arm: `cmakelibmpfr-dev` looks like a typo.

Also, please remove clang for now. Let's make it work with one compiler, then extend it.

.travis.yml
```diff
@@ -4,21 +4,14 @@ dist: trusty
 group: edge
 
 before_install:
```
shibatch: You also need libomp-dev if you compile sleef with clang.
Also note that clang version 3.9 and later are only supported.

.travis.yml
```diff
@@ -4,11 +4,18 @@ dist: trusty
...
+    - sudo add-apt-repository -y ppa:adrozdoff/cmake
     - sudo apt-get -qq update
-    - sudo apt-get install -y libmpfr-dev libgomp1 gcc
-    
+    - sudo apt-get install -y cmakelibmpfr-dev libgomp1 gcc
```
fpetrogalli-arm: `cmakelibmpfr-dev` is a typo.

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+target_compile_options(${TARGET_MKRENAME}
+  PRIVATE ${FLAGS_BASIC_C} ${FLAGS_WALL})
+
+# Helper executable: dispatcher
+add_executable(${TARGET_MKDISP} mkdisp.c)
+target_compile_options(${TARGET_MKDISP}
+  PRIVATE ${FLAGS_BASIC_C} ${FLAGS_WALL})
+
+# TODO: set the parameters for mkrename run
+set(MKRENAME_PARAMS "")
```
fpetrogalli-arm: I believe this is needed only for the SIMD implementation - can you update your comment?

```
set the parameters for mkrename runs for the SIMD implementation
```

src/libm-tester/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+# Target 'test-libm' runs the libm tests
+add_custom_target(${TARGET_TEST}
+  COMMAND ./tester ./iut
+  WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+add_dependencies(${TARGET_TEST} ${TARGET_TESTER} ${TARGET_IUT} ${TARGET_LIBSLEEF})
+
+# Install test executables in the bin dir of the build directory
+install(TARGETS ${TARGET_IUT} ${TARGET_TESTER}
+  DESTINATION bin)
```
fpetrogalli-arm: Please make the CMAKE_BINARY_DIR explicit

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+# Add custom command to install the rename header under /include
+add_custom_command(
+  OUTPUT ${CMAKE_BINARY_DIR}/include/rename.h
+  COMMAND ${CMAKE_COMMAND} -E copy
+    ${CMAKE_CURRENT_SOURCE_DIR}/rename.h ${CMAKE_BINARY_DIR}/include/rename.h
+  DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/rename.h)
+
+# Build main library
+add_library(${TARGET_LIBSLEEF}
+  SHARED sleefdp.c sleefsp.c ${CMAKE_BINARY_DIR}/include/sleef.h ${CMAKE_BINARY_DIR}/include/rename.h)
```
fpetrogalli-arm: I think that `rename.h` should not be listed as a source file but via an include path, as it is used as follows:

```
#ifdef DORENAME
#include "rename.h"
#endif
```

src/libm-tester/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+add_dependencies(${TARGET_IUT} install_headers)
+target_link_libraries(${TARGET_IUT} ${LIB_SLEEF} ${EXTRA_LIBS})
+target_compile_options(${TARGET_IUT}
+  PRIVATE ${FLAGS_BASIC_C} ${FLAGS_WALL} ${FLAGS_STRICTMATH})
+
+# Compiler executable 'tester'
+add_executable(${TARGET_TESTER} tester.c testerutil.c)
+target_link_libraries(${TARGET_TESTER} ${EXTRA_LIBS})
+target_compile_definitions(${TARGET_TESTER}
+  PUBLIC USEMPFR=1)
```
fpetrogalli-arm: Why PUBLIC? Couldn't this be a PRIVATE definition?

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
+# Helper executable: generates parts of the sleef header file
+add_executable(${TARGET_MKRENAME} mkrename.c)
+target_compile_options(${TARGET_MKRENAME}
+  PRIVATE ${FLAGS_BASIC_C} ${FLAGS_WALL})
+
+# Helper executable: dispatcher
+add_executable(${TARGET_MKDISP} mkdisp.c)
```
fpetrogalli-arm: I believe this executable is not used anywhere in this current pull request. Please remove it.

CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+set(TARGET_LIBSLEEF "sleef")
+# Generates executable files for running the test suite; defined in src/libm-tester/CMakeLists.txt via command add_executable
+set(TARGET_IUT "iut")
+set(TARGET_TESTER "tester")
+# Runs the test suite; depends on targets: iut, tester; defined in src/libm-tester/CMakeLists.txt via command add_custom_target
+set(TARGET_TEST "test-libm")
+# Generates the helper executable file mkrename needed to write the sleef header file
+set(TARGET_MKRENAME "mkrename")
+# Generates the helper executable file mkdisp
+set(TARGET_MKDISP "mkdisp")
```
fpetrogalli-arm: Please remove (comment included)

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+  SHARED sleefdp.c sleefsp.c ${CMAKE_BINARY_DIR}/include/sleef.h ${CMAKE_BINARY_DIR}/include/rename.h)
+
+if(ENABLE_LONG_DOUBLE)
+  target_sources(${TARGET_LIBSLEEF} PRIVATE sleefld.c)
+endif(ENABLE_LONG_DOUBLE)
+
+if(ENABLE_FLOAT128)
+  target_sources(${TARGET_LIBSLEEF} PRIVATE sleefqp.c)
+  target_compile_definitions(${TARGET_LIBSLEEF}
+    PRIVATE ENABLEFLOAT128=1)
```
shibatch: ENABLEFLOAT128 should be replaced with COMPILER_SUPPORTS_FLOAT128.

src/libm/CMakeLists.txt
diaena: Done.

src/libm/CMakeLists.txt
diaena: Done.

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+  PRIVATE ${FLAGS_BASIC_C} ${FLAGS_WALL})
+
+# TODO: set the parameters for mkrename run
+set(MKRENAME_PARAMS "")
+
+configure_file(
+  ${CMAKE_CURRENT_SOURCE_DIR}/GenerateSleefHeader.cmake.in
+  ${CMAKE_CURRENT_BINARY_DIR}/GenerateSleefHeader.cmake @ONLY)
+
+# Add sleeh header target
```
diaena: Done.

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+add_custom_command(
+  OUTPUT ${CMAKE_BINARY_DIR}/include/sleef.h
+  COMMAND ${CMAKE_COMMAND} -P
+    ${CMAKE_CURRENT_BINARY_DIR}/GenerateSleefHeader.cmake
+  DEPENDS
+    ${CMAKE_CURRENT_SOURCE_DIR}/sleeflibm.h.org
+    ${CMAKE_CURRENT_BINARY_DIR}/GenerateSleefHeader.cmake)
+
+# Install header files in the include directory
+add_custom_target(install_headers ALL
```
diaena: This is indeed not very clear and could be quite confusing for someone picking up SLEEF development. 

Currently, all header files are put into the ${CMAKE_BUILD_DIR}/include and that is included via -I when compiling. I am doing this to avoid having to specify relative paths in the cmake files.  However, at installation time, only sleef.h is installed under ${CMAKE_INSTALL_DIR}/include.

I think we need to have a conversation about the header files sooner or later. I have a suggestion I am still working on and will submit a pull request after the SIMD version is building. But for now I think it's okay not to worry about the tree structure.

src/libm/CMakeLists.txt
fpetrogalli-arm: @shibatch, I believe you refer at `ENABLE_FLOAT128`? If so, @diaena, please also change `ENABLE_LONG_DOUBLE` to `COMPILER_SUPPORTS_LONG_DOUBLE`, as those are the naming conventions we agreed.

src/libm/CMakeLists.txt
fpetrogalli-arm: Sounds good to me - mark it as a TODO comment.

CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
+# See doc/build-with-cmake.md for instructions on how to build Sleef.
+cmake_minimum_required(VERSION 3.4.3)
+
+set(SLEEF_VERSION_MAJOR 3)
+set(SLEEF_VERSION_MINOR 1)
+set(SLEEF_SOVERSION ${SLEEF_VERSION_MAJOR})
+
+project(SLEEF
+	VERSION ${SLEEF_VERSION_MAJOR}.${SLEEF_VERSION_MINOR}
+	LANGUAGES C CXX)
```
diaena: I thought we did. Done.

src/libm/CMakeLists.txt
shibatch: At this time, the only generated header that has to be installed is sleef.h, so I think there is not much confusion. If we will have more generated headers that have to be installed, one way to make it tidy would be separating include directories.

diaena: Sorry, I meant confusion introduced by my workaround. I have modified my code to emphasise what I've meant by avoiding relative paths includes.

src/libm/CMakeLists.txt
diaena: Indeed, that was added there as dependency. To be more specific, I needed to tell CMake that the command with the output '/include/rename.h' needs to be executed before attempting to compile the library.

src/libm/CMakeLists.txt
diaena: No problem, done.

src/libm/CMakeLists.txt
fpetrogalli-arm: Ah - make sense, so it is a target. Can you list it as a `TARGET_RENAME_H` in the main cmake configuration file, like you have done for other targets? At this point also `${CMAKE_BINARY_DIR}/include/sleef.h` should become `TARGET_SLEEF_H`.

src/libm-tester/CMakeLists.txt
diaena: Good point. Done.

src/libm-tester/CMakeLists.txt
diaena: Right, so DESTINATION is the install directory, not the CMake build directory. It's the one you specify at configuration time via CMAKE_INSTALL_PREFIX and is by default /usr/local. 

I realise now that we might not want to install the test executables. Will remove.

diaena: Reference for CMake docs:
the install DESTINATION: https://cmake.org/cmake/help/v3.0/command/install.html

fpetrogalli-arm: yes, we don't want to install the tests suite.

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+
+# Always compile sleef with -ffp-contract and log at configuration time
+if(SLEEF_SHOW_CONFIG)
+  message(STATUS "Using option `${FLAGS_STRICTMATH}` to compile libsleef")
+endif(SLEEF_SHOW_CONFIG)
+
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/sleef.h
+  DESTINATION include)
+install(TARGETS ${TARGET_LIBSLEEF}
+  DESTINATION lib)
```
diaena: Done and explained. I don't think this needs explaining in the code as I would just explain cmake mechanisms.

.travis.yml
```diff
@@ -1,14 +1,26 @@
...
     - gcc
 
-script: make test
+before_script:
+  - mkdir sleef.build
+  - cd sleef.build
+  - cmake -DCMAKE_BUILD_TYPE=Release ..
+
+script:
+  - make test-libm VERBOSE=1
```
fpetrogalli-arm: I think you can remove `VERBOSE=1` from here.

Also, add a `make install` after the testing to make sure that also testing ends correctly.

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/sleeflibm.h.org
+    ${CMAKE_CURRENT_BINARY_DIR}/GenerateSleefHeader.cmake)
+
+# Add target for generating the sleef header
+add_custom_target(generate-header ALL
+  DEPENDS ${CMAKE_BINARY_DIR}/include/sleef.h)
+
+# Build main library
+add_library(${TARGET_LIBSLEEF}
+  SHARED sleefdp.c sleefsp.c ${CMAKE_BINARY_DIR}/include/sleef.h)
```
fpetrogalli-arm: Please use a target at top level for `${CMAKE_BINARY_DIR}/include/sleef.h`

src/libm-tester/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+link_directories(${CMAKE_BINARY_DIR}/lib)
+include_directories(${CMAKE_BINARY_DIR}/include)
+
+# This is why a conversation about headers is needed.
+# TODO: Find way to avoid relative includes.
+include_directories(../libm)
```
fpetrogalli-arm: can't this be something like `${CMAKE_SOURCE_DIR}/src/libm`?

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+set_target_properties(${TARGET_LIBSLEEF} PROPERTIES
+  VERSION ${SLEEF_VERSION_MAJOR}.${SLEEF_VERSION_MINOR}
+  SOVERSION ${SLEEF_SOVERSION})
+
+# Always compile sleef with -ffp-contract and log at configuration time
+if(SLEEF_SHOW_CONFIG)
+  message(STATUS "Using option `${FLAGS_STRICTMATH}` to compile libsleef")
+endif(SLEEF_SHOW_CONFIG)
+
+install(FILES ${CMAKE_BINARY_DIR}/include/sleef.h
```
fpetrogalli-arm: Please define a target in the top level folder for `${CMAKE_BINARY_DIR}/include/sleef.h`

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+# TODO: set the parameters for mkrename run for the SIMD implementation
+set(MKRENAME_PARAMS "")
+
+configure_file(
+  ${CMAKE_CURRENT_SOURCE_DIR}/GenerateSleefHeader.cmake.in
+  ${CMAKE_CURRENT_BINARY_DIR}/GenerateSleefHeader.cmake @ONLY)
+
+# Add custom command to generate the sleef header at build time
+add_custom_command(
+  OUTPUT ${CMAKE_BINARY_DIR}/include/sleef.h
```
fpetrogalli-arm: Please define a target in the top level folder for `${CMAKE_BINARY_DIR}/include/sleef.h`.

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+  COMMAND ${CMAKE_COMMAND} -P
+    ${CMAKE_CURRENT_BINARY_DIR}/GenerateSleefHeader.cmake
+  DEPENDS
+    ${CMAKE_CURRENT_SOURCE_DIR}/sleeflibm.h.org
+    ${CMAKE_CURRENT_BINARY_DIR}/GenerateSleefHeader.cmake)
+
+# Add target for generating the sleef header
+add_custom_target(generate-header ALL
+  DEPENDS ${CMAKE_BINARY_DIR}/include/sleef.h)
```
fpetrogalli-arm: Please add a target in the top level folder for `${CMAKE_BINARY_DIR}/include/sleef.h`.

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+add_custom_command(
+  OUTPUT ${CMAKE_BINARY_DIR}/include/sleef.h
+  COMMAND ${CMAKE_COMMAND} -P
+    ${CMAKE_CURRENT_BINARY_DIR}/GenerateSleefHeader.cmake
+  DEPENDS
+    ${CMAKE_CURRENT_SOURCE_DIR}/sleeflibm.h.org
+    ${CMAKE_CURRENT_BINARY_DIR}/GenerateSleefHeader.cmake)
+
+# Add target for generating the sleef header
+add_custom_target(generate-header ALL
```
fpetrogalli-arm: generate-header doesn't seem to be mentioned anywhere. Am I missing something?

src/libm/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+    PRIVATE ENABLEFLOAT128=1)
+endif(COMPILER_SUPPORTS_FLOAT128)
+
+target_compile_definitions(${TARGET_LIBSLEEF}
+  PRIVATE DORENAME=1)
+target_compile_options(${TARGET_LIBSLEEF}
+  PUBLIC ${FLAGS_WALL} ${FLAGS_STRICTMATH})
+# Selects C99 with standard gnu99 (add -std=gnu99) for gcc
+target_compile_features(${TARGET_LIBSLEEF}
+  PUBLIC c_variadic_macros)
```
fpetrogalli-arm: what is this? why not just add `-std=gnu99`? is it a compiler independent representation in cmake of the feature activated with `-std=gnu99` in gcc?

src/libm/CMakeLists.txt
diaena: You can see it mentioned at configuration time?

src/libm/CMakeLists.txt
diaena: Done.

.travis.yml
diaena: Done.

src/libm/CMakeLists.txt
diaena: I have changed this after learning about some CMake best practices. It's considered bad practice to specify flags such as -std=c++11 or -std=gnu99. Apparently, it's best to leave that job to CMake. You should specify a compile feature that you need and then CMake will automatically add the -std flags.

CMake 3.8 introduces standard levels so you wouldn't have to specify a granular feature. 
i.e. you could do: 
`target_compiler_features(mytarget PUBLIC c_std_99`

Ref: https://cmake.org/cmake/help/v3.8/prop_gbl/CMAKE_C_KNOWN_FEATURES.html

src/libm/CMakeLists.txt
diaena: it's tricky because that's a custom command output as opposed to a target.

src/libm-tester/CMakeLists.txt
diaena: That would assume that SLEEF is the root project. In other words, you can't integrate this Cmake project with other Cmake projects (such as LLVM).

src/libm/CMakeLists.txt
fpetrogalli-arm: I think this is not needed. The custom_command dependency is already specified in all targets when you use `${CMAKE_BINARY_DIR}/include/sleef.h`. Please remove.

src/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+include_directories("common")
```
fpetrogalli-arm: I think this can be removed, as it doesn't seem to be used anywhere.

diaena: misc.h included by sleefdp.c

src/libm/CMakeLists.txt
fpetrogalli-arm: Thanks for explaining it, I agree with you now on the "best leave it to cmake".

PR #41, opened 2017-07-07
Flush stdin / stdout in the tester.

fpetrogalli-arm:
This patch forces the fflushing of `stdin` and `stdout` in `tester.c` to prevent CPU feature detection failures.


fpetrogalli-arm:
I have added some comments related to the fix.

shibatch:
I think bugfix and bug workaround are different.

shibatch:
Oops, I meant approving the pull request.

PR #42, opened 2017-07-17
This patch updates the documentation. The updated documentation can b…

shibatch:
…e viewed at http://sleef.org/

It also changes minor version numbers in DLL, and update files related to debian packaging.

I am planning to release version 3.1 after this pull request is approved.


fpetrogalli-arm:
Hello @shibatch , it is very good that you have updated the documentation.

I only have a couple of comments:

1. why do you document `aarch64` and `aarch32` functionalities in `aarch64.xhtml` and `aarch32.xhtml`? Aren't those function descriptions valid also for other targets? If so, givent thet the library is target independent, why not describe the generic functions `xsin` , `xatan2f` and so on?
2. I think it would be better to avoid html in document sources, as it makes difficult to review and maintain. You could use markdown - the main advantages would be:
   a. it would make the formatted output directly visible on github, as for example https://github.com/diaena/sleef/blob/cff3d60176462ac4a1ff38c6764f33f2e4dea760/doc/build-with-cmake.md
   b. markdown could be easily converted to html for the website with pandoc, (http://www.pandoc.org).

Thank you,

Francesco

shibatch:
For 1., currently there is almost no description for each function. They are just pointing the scalar functions. 

For 2., those documents will be changed to doxygen-based ones before release of 3.2. At this point, it was quicker to directly write html-based documents since I need experiments to migrate to doxygen or other tools.

fpetrogalli-arm:
I see - it make sense then. Thank you!

PR #43, opened 2017-07-26
Add clang testing in the CI.

diaena:
There will be subsequent commits. I do not know what else this changes requires yet as there is no way of triggering Travis before hand.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,29 @@
 language: c
 
-# Default job
+# Default linux jobs
 os: linux
 sudo: required
 dist: trusty
-# Should spawn a linux-gcc job and a linux-clang job
-compiler: 
+compiler:
   - gcc
   - clang
 
-# Aditional testing on OSX with default compiler: should spawn one job
+# Include osx jobs
 matrix:
   include:
     - os: osx
-      env: COMPILER=clang
+      osx_image: xcode8
+      compiler: gcc
+      env:
+        - MATRIX_EVAL="brew install gcc@6 && CC=gcc-6" # install and use gcc
+    - os: osx
+      compiler: clang # use default apple clang
 
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo add-apt-repository -y ppa:adrozdoff/cmake; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get -qq update                       ; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install -y cmake libmpfr-dev     ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx"   ]]; then eval "${MATRIX_EVAL}"                         ; fi
 
 before_script:
   - mkdir sleef.build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,13 @@
 language: c
 
-matrix:
-  include:
-    - os: linux
-      sudo: required
-      dist: trusty
-      group: edge
-    - os: osx
+os:
+  - linux
+  - osx
 
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo add-apt-repository -y ppa:adrozdoff/cmake        ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get -qq update                               ; fi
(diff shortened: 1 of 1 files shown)


shibatch:
I don't fully understand your situation, but the check is failing because travis is trying to build on precise. You can just do checking on trusty.

diaena:
I understand it's failing because it's trying to build on Precise, but I don't understand why it's trying to build on Precise. I am specifically setting the distribution to "trusty" in the configuration file. And only two out of four jobs use the Precise instead of Trusty.

diaena:
Also, it there a reason for which we are using trusty in group: edge?

shibatch:
The reason for specifying group:edge is that that was once required to build on Trusty. You don't need it anymore.
I am not an expert of travis, but I guess the reason why it is trying to build on Precise is that you specify include. The default build environment is still Precise.

.travis.yml
```diff
@@ -1,20 +1,24 @@
...
+# Default job
+os: linux
+sudo: required
+dist: trusty
+# Should spawn a linux-gcc job and a linux-clang job
+compiler: 
+  - gcc
+  - clang
+
+# Aditional testing on OSX with default compiler: should spawn one job
```
fpetrogalli-arm: Typo

fpetrogalli-arm: Typo

.travis.yml
```diff
@@ -1,20 +1,24 @@
...
 matrix:
   include:
-    - os: linux
-      sudo: required
-      dist: trusty
-      group: edge
     - os: osx
+      env: COMPILER=clang
```
fpetrogalli-arm: Is this saying that we test on osx only with clang?

shibatch: @diaena Please carefully read the following page.
https://docs.travis-ci.com/user/languages/c/

Do you have Intel compiler installed at your environment?
Support for Intel compiler was added following requests from the users.
Even though they can use SVML, they still want to use SLEEF.

fpetrogalli-arm:
It seems that Intel compiler can be installed on travis-ci, so let's try that route. This github issue provides some solutions: https://github.com/travis-ci/travis-ci/issues/4604

The compiler can be installed via `apt-get` [1]. Alternatively, someone provided a script [2]. Please add this compiler on a different pull request, and only for the current scalar implementation, not the SIMD one.


[1] https://software.intel.com/en-us/articles/installing-intel-parallel-studio-xe-from-apt-repository
[2] https://github.com/nemequ/icc-travis

.travis.yml
diaena: @fpetrogalli-arm Sorry, no, that was just me trying to understand what the env does and what it affects.
@shibatch I think it is generally considered bad practice to specify commands (or install packages) if it's not required. 

I will add gcc on OSX in my next patch, so we would test the following:

* Sleef on Linux with gcc-v4.8.4 (default gcc on Trusty)
* Sleef on Linux with clang-v3.5.0 (default clang on Trusty)
* Sleef on OSX with  Apple LLVM version 7.3.0 (clang-703.0.31) (default on OSX)
* Sleef on OSC with gcc-6 (needs installing) -- Are we okay with using latest version?

shibatch: It would be a matter of preference, but I would tolerate small amount of deviance if it helps other things like ease of development.

gcc-4 has a problem with long double. clang earlier than v3.9 has problems with openmp.
gcc-6 should work without any problem.
gcc-7.1 doesn't compile DFT correctly. We need to remove -ffast-math for gcc-7.1.

diaena:
I know this has been approved, but I am not not yet merging it because I have double-checked and the OSX jobs are overwriting my CC variable and therefore both jobs build with Apple Clang. I am now trying to find a workaround this other Travis bug. It shouldn't be the case that the CC variable is re-exported after you evaluate the Matrix because the documentation specifically uses this as an example. I will submit another patch and hopefully this time we get the compilers we actually want to test.

diaena:
For reference, this is what I'm talking about: 
image

And in this case, gcc is just a frontend for LLVM. The same is happening here: 

image

shibatch:
I don't know but can't you test that kind of things on your local environment?
Do you run cmake on your computer before testing on travis?

diaena:
Of course I do, but this is not related to cmake or SLEEF. This is Travis CI not selecting the compiler I am telling it to select.

fpetrogalli-arm:
It is not clear to me which compiler is being used from the logs... Maybe it is time to turn on `VERBOSE=1` in the `make libm-test` invocation to see what is going on? To speed up things, you might want to turn off the testing and just run `make VERBOSE=1` in this development branch.

diaena:
@fpetrogalli-arm To see the cmake configuration output, click on the arrow on line 1401 : https://travis-ci.org/shibatch/sleef/jobs/258097159

PR #45, opened 2017-08-08
Add symbols for the SSE4 and AVX vector architectures + tests.

diaena:


Configure.cmake
```diff
@@ -42,6 +45,28 @@ CHECK_C_SOURCE_COMPILES("
...
+  #if defined(_MSC_VER)
+  #include <intrin.h>
+  #else
+  #include <x86intrin.h>
+  #endif
+  int main() {
+    __m128d r = _mm_floor_sd(_mm_set1_pd(1), _mm_set1_pd(2)); }"
+  COMPILER_SUPPORTS_SSE4)
+
+set (CMAKE_REQUIRED_FLAGS ${FLAGS_ENABLE_AVX})
```
fpetrogalli-arm: Is this what it is required to compile AVX intrinsics? If so, I see you are using it also for SSE2 and SSE4.
Does this variable get reset after the checks are done? Or is it carried all through the cmake configuration fase?

cmake/Scripts/GenerateSleefHeader.cmake
```diff
@@ -19,23 +10,31 @@ if(OPTION_SHOW_CONFIG)
 endif(OPTION_SHOW_CONFIG)
 
 # Generate header for x86
+  set(run1 2 4 __m128d __m128 __m128i __m128i __SSE2__)
```
fpetrogalli-arm: May I ask you to give meaningful names instead of `run*`?

cmake/Scripts/GenerateSleefHeader.cmake
```diff
@@ -19,23 +10,31 @@ if(OPTION_SHOW_CONFIG)
...
+
 # TODO: Change condition of generation to use a COMPILER_SUPPORTS_* flag
 if(SLEEF_ARCH_X86)
-  list(APPEND MKRENAME_PARAMS_LIST
-    "2 4 __m128d __m128 __m128i __m128i __SSE2__")
-  list(APPEND MKRENAME_PARAMS_LIST
-    "2 4 __m128d __m128 __m128i __m128i __SSE2__ sse2")
-  list(APPEND MKRENAME_PARAMS_LIST
-    "2 4 __m128d __m128 __m128i __m128i __SSE2__ sse4")
+  list(APPEND PARAMS_POINTER_LIST run1 run2 run3 run4 run5 run6 run7 run8 run9)
```
fpetrogalli-arm: I don't understand how this is working. Why is the list carrying 9 elements instead of the single values of each of the runs?

cmake/Scripts/GenerateSleefHeader.cmake
```diff
@@ -19,23 +10,31 @@ if(OPTION_SHOW_CONFIG)
...
 endif()
 
-foreach(params_set ${MKRENAME_PARAMS_LIST})
-  run_mkrename(${params_set} RESULT_VARIABLE)
-  set(OUTPUT "${OUTPUT}\n${RESULT_VARIABLE}")
+foreach(params_set ${PARAMS_POINTER_LIST})
```
fpetrogalli-arm: I am sure I am missing something, but how comes that the param_set is grouping the params correctly? Could you explain this in the comments?

cmake/Scripts/GenerateSleefHeader.cmake
```diff
@@ -19,23 +10,31 @@ if(OPTION_SHOW_CONFIG)
...
+  execute_process(
+      COMMAND ${LOCATION_RUNTIME_DIR}/${TARGET_MKRENAME} ${${params_set}}
```
fpetrogalli-arm: If you define run1 to be `set(run1 "2 4 __m128d __m128 __m128i __m128i __SSE2__")`, would this allow you to replace `${${params_set}}` with `${params_set}`?

src/libm-tester/CMakeLists.txt
```diff
@@ -27,6 +27,28 @@ if(COMPILER_SUPPORTS_SSE2)
...
 
+if(COMPILER_SUPPORTS_SSE4)
+  add_executable(${TARGET_IUTSSE4} iutsimd.c iutsimdmain.c testerutil.c)
+  target_compile_options(${TARGET_IUTSSE4}
+    PRIVATE ${FLAGS_ENABLE_SSE4})
+  target_compile_definitions(${TARGET_IUTSSE4}
+    PRIVATE ENABLE_SSE4=1)
+  target_link_libraries(${TARGET_IUTSSE4} ${LIB_SLEEF} ${EXTRA_LIBS})
+
+  list(APPEND IUT_LIST ${TARGET_IUTSSE4})
```
fpetrogalli-arm: This, and all the similar lines appending testers to the IU_LIST variable, should be guarded by different variables other then `COMPILER_SUPPORT_*`. The fact is that a target might not have the vector extension that the compiler has been able to compile, so that the actual `IUT_LIST` should be updated by a `TARGET_RUNS_*`, which should be produced out  of `try_run` instead of `try_compile`.

The code is working as it is because the `tester` does not report a failure if the `iut*` that is being executed is not supported by the target. It is fine for me to leave this as it is, but I would like to hear @shibatch 's opinion.

shibatch: I would prefer not to complicate cmakefile too much. So, I think the current implementation is okay.

Configure.cmake
shibatch: AVX is not available for Cygwin because its ABI does not have support yet. I think it is okay to add detection for that here. It does not make much sense to do detection using CHECK_C_SOURCE_COMPILES here, since all the supported compiler should compile this.

One way to detect if there is ABI support is to check if the compiler is Cygwin.
Another way is to check if the compiled program runs correctly.
The following program runs normally on linux, but segfaults on Cygwin.

#include <stdio.h>
#include <x86intrin.h>

__m256d check() {
  return _mm256_set1_pd(1);
}

int main() {
  __m256d v = check();
  double a[4];
  _mm256_storeu_pd(a, v);
  printf("%g %g %g %g\n", a[0], a[1], a[2], a[3]);
}

If we think about cross-compiling, detecting Cygwin compiler would be better.
If we think about future possibility for this problem to be fixed, checking if the program runs would be better.

fpetrogalli-arm: This would require to add the `try_run` checks that I mentioned in https://github.com/shibatch/sleef/pull/45#discussion_r132029105

I am fine doing that, only I would like to avoid as much as possible checks in the form "if compiler is [whatever]", as I prefer to ask questions in the form "does compiler supports [feature]" or "does target can run [feature]", as they are target and compiler independent.

cmake/Scripts/GenerateSleefHeader.cmake
diaena: I actually tried to come up with something, but it's just mkrename run parameters really and they're not used anywhere else. Also there is no need to differentiate between the variables or anything like that.

fpetrogalli-arm: > Also there is no need to differentiate between the variables or anything like that.

What do you mean?

cmake/Scripts/GenerateSleefHeader.cmake
diaena: No.

Variable _params_set_ is a cmake list (contains the _names_ of the mkrename runs, ie. list of parameters). 
Therefore, dereferencing the variable gives you the values:
`${params_set} = run1`
Furthermore, double-dereferencing gives you the values of the lists in your list of (list)names.
`${${params_set}} = 2 4 __m128d __m128 __m128i __m128i __SSE2__`
NOTE: This value is a list, not a string. The reason I need a list is because that's what I need to pass to mkrename, a list of parameters as opposed to one string.

This is basically a pointer in cmake. And I needed it to implement a list of lists.

cmake/Scripts/GenerateSleefHeader.cmake
diaena: See comment below.

cmake/Scripts/GenerateSleefHeader.cmake
fpetrogalli-arm: Nice explanation - add it as a comment to the sources when you use ${${parames_set}} for the first time.

cmake/Scripts/GenerateSleefHeader.cmake
diaena: I mean, you do not need to refer to a specific set of parameters anywhere. You either call all the runs in sequence or you don't. 
There is no context in which you need to which set is run6 and which set is run9.

Configure.cmake
diaena: I agree that we should check for features as opposed to platform specifications. In my opinion, default detection should be done for the host system. Once we add cross-compilation support, we worry about target system specifications and such.

cmake/Scripts/GenerateSleefHeader.cmake
diaena: Done

Configure.cmake
fpetrogalli-arm: Can we leave this as it was originally? I don't like the solution of using `CHECK_C_SOURCE_RUNS` to set `COMPILER_SUPPORT_*` as much as I don't like checking if the compiler is Cygwin.

If we use the original version to use `CHECK_C_SOURCE_COMPILES` to set `CMAKE_SUPPORTS_AVX` on Cygwin, we would end up building a library that provides symbols (the AVX ones) that the system (Cygwin) is not able to run? Is that a bad thing? @shibatch what do you think? I don't see why this is a bad thing, as we are already building the library for AVX512  on linux x86 even if the target system does not support it. Am I wrong?

diaena: I think there is no right or wrong, but it's a matter of what we decide. From what I understand, the question is: 
_Should the library build symbols that the native host does not support?_

If we currently build AVX512 on systems that can't run it, then I see no problem in doing the same for AVX on Cygwin. 
There is the other take that maybe it's not a good idea to provide vector symbols for architectures that the _native_ host system does not support, but only that for cross-compiled builds. 

I think we need to decide and be consistent, but I might have misunderstood and need some clarification? @shibatch

shibatch: Yeah, there is basically no harm if the AVX functions actually do not work. So, it is a matter of preference. It is okay that AVX functions are compiled into the Cygwin version of the library. It is better to explain this in the document, though.

diaena: Ok then, I am rolling back to do a try_compile and will add a little comment to document the Cygwin issue.

diaena:
@shibatch  Are you happy to merge this?

fpetrogalli-arm:
@shibatch, Diana has another couple of patches that depend on this one. I don't see anything controversial in what she wrote and in your comments, so we will not wait for your formal approval to submit this. We will address any of your concerns in a later patch if you have any.

@diaena, please "Squash and merge" this patch.

PR #54, opened 2017-08-14
[Experimental] Add support for the x86 vectors: FMA4, AVX2, AVX2128, AVX512F

diaena:


diaena:
Right, so, in the previous pull request (that I have accidentally merged in) you said to make sure I am testing the AVX512F code. I have a couple of questions: 

1. Are we testing this code in the current CI?
2. Is there a better way to detect compiler support for AVX512F? (My build currently fails.)

I'd greatly appreciate some help with this, thanks!

shibatch:
Even AVX2 is not supported on the travis servers. So, it is not tested. GCC4 does not support AVX512F.
For some reason, the current detection method succeeds on Apple clang but the AVX512F code does not compile.  I don't know which clang version this Apple clang corresponds to.
Since it is saying _mm512_loadu_si512 and some other symbols are not defined, we should use these symbols to detect AVX512F support.

diaena:
Thanks for your reply. I think all AVX512F code will compile with Apple Clang. The errors are popping up in the helper header. I am pushing the other detection code, but the build still fails.

shibatch:
Build is passing. I think it is now working as we expected.
What is failing is the build on AppVeyor, which I added yesterday.

diaena:
I've also made sure AVX2 is passing. Tested it locally on my Mac.
Francesco is away on holiday so I don't expect a review from him any time soon. Is it okay to merge this in?

PR #55, opened 2017-08-16
Refactor

diaena:
* Use macros to compile objects and test executables
* Avoid repeating the same code for each vector extension
* Add global list variable with the extensions supported by SLEEF

### DIFF
--- a/Configure.cmake
+++ b/Configure.cmake
@@ -71,8 +71,8 @@ CHECK_C_SOURCE_COMPILES("
   #include <x86intrin.h>
   #endif
   int main() {
-    __m128d r = _mm_mul_pd(_mm_set1_pd(1), _mm_set1_pd(2));
-  }" COMPILER_SUPPORTS_SSE2)
+    __m128d r = _mm_mul_pd(_mm_set1_pd(1), _mm_set1_pd(2)); }"
+  COMPILER_SUPPORTS_SSE2)
 
 set (CMAKE_REQUIRED_FLAGS ${FLAGS_ENABLE_SSE4})
 CHECK_C_SOURCE_COMPILES("
@@ -82,8 +82,8 @@ CHECK_C_SOURCE_COMPILES("
   #include <x86intrin.h>
   #endif
   int main() {
-    __m128d r = _mm_floor_sd(_mm_set1_pd(1), _mm_set1_pd(2));
-  }" COMPILER_SUPPORTS_SSE4)
+    __m128d r = _mm_floor_sd(_mm_set1_pd(1), _mm_set1_pd(2)); }"
+  COMPILER_SUPPORTS_SSE4)
 
 set (CMAKE_REQUIRED_FLAGS ${FLAGS_ENABLE_AVX})
 CHECK_C_SOURCE_COMPILES("
@@ -104,8 +104,8 @@ CHECK_C_SOURCE_COMPILES("
   #include <x86intrin.h>
   #endif
   int main() {
-    __m256d r = _mm256_macc_pd(_mm256_set1_pd(1), _mm256_set1_pd(2), _mm256_set1_pd(3));
-  }" COMPILER_SUPPORTS_FMA4)
+    __m256d r = _mm256_macc_pd(_mm256_set1_pd(1), _mm256_set1_pd(2), _mm256_set1_pd(3)); }"
+  COMPILER_SUPPORTS_FMA4)
 
 set (CMAKE_REQUIRED_FLAGS ${FLAGS_ENABLE_AVX2})
 CHECK_C_SOURCE_COMPILES("
@@ -115,8 +115,8 @@ CHECK_C_SOURCE_COMPILES("
   #include <x86intrin.h>
   #endif
   int main() {
-    __m256i r = _mm256_abs_epi32(_mm256_set1_epi32(1));
-  }" COMPILER_SUPPORTS_AVX2)
+    __m256i r = _mm256_abs_epi32(_mm256_set1_epi32(1)); }"
+  COMPILER_SUPPORTS_AVX2)
 
 set (CMAKE_REQUIRED_FLAGS ${FLAGS_ENABLE_AVX512F})
 CHECK_C_SOURCE_COMPILES("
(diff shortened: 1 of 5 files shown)


diaena:
This change compresses the code quite a bit and it makes it much easier for me to build the gnuabi version.

src/libm-tester/CMakeLists.txt
```diff
@@ -16,94 +17,31 @@ target_link_libraries(${TARGET_IUT} ${LIB_SLEEF} ${EXTRA_LIBS})
...
+    target_link_libraries(${TARGET_IUT${SIMD}} ${LIB_SLEEF} ${EXTRA_LIBS})
+
+    list(APPEND IUT_LIST ${TARGET_IUT${SIMD}})
+  endif(COMPILER_SUPPORTS_${SIMD})
+endmacro(test_extension)
+
+set(IUT_SRC iutsimd.c iutsimdmain.c testerutil)
+
+foreach(SIMD ${SLEEF_SUPPORTED_EXTENSIONS})
+  if(SIMD STREQUAL "ADVSIMD")
```
fpetrogalli-arm: Please build iutadvsimd like this for now, and create an issue saying that we need to build iutadvsimd like the other iut* executables.

Fixing this now would require changes in the c/h sources and in the makefiles that might break the makefiles builds. I'd rather leave it as it is.

fpetrogalli-arm: Just add a comment with a link to the issue for fixing this.

Configure.cmake
```diff
@@ -51,8 +71,8 @@ CHECK_C_SOURCE_COMPILES("
   #include <x86intrin.h>
   #endif
   int main() {
-    __m128d r = _mm_mul_pd(_mm_set1_pd(1), _mm_set1_pd(2)); }"
-  COMPILER_SUPPORTS_SSE2)
+    __m128d r = _mm_mul_pd(_mm_set1_pd(1), _mm_set1_pd(2));
```
fpetrogalli-arm: This is not related to the refactoring, please revert

Configure.cmake
```diff
@@ -62,8 +82,8 @@ CHECK_C_SOURCE_COMPILES("
   #include <x86intrin.h>
   #endif
   int main() {
-    __m128d r = _mm_floor_sd(_mm_set1_pd(1), _mm_set1_pd(2)); }"
-  COMPILER_SUPPORTS_SSE4)
+    __m128d r = _mm_floor_sd(_mm_set1_pd(1), _mm_set1_pd(2));
```
fpetrogalli-arm: This  not related to the refactoring, please revet

Configure.cmake
```diff
@@ -78,49 +98,46 @@ CHECK_C_SOURCE_COMPILES("
...
-#if defined(_MSC_VER)
-#include <intrin.h>
-#else
-#include <x86intrin.h>
-#endif
-int main() {
-  __m256d r = _mm256_macc_pd(_mm256_set1_pd(1), _mm256_set1_pd(2), _mm256_set1_pd(3));
-}" COMPILER_SUPPORTS_FMA4)
+  #if defined(_MSC_VER)
+  #include <intrin.h>
```
fpetrogalli-arm: This is not related to the refactoring please revert

Configure.cmake
```diff
@@ -78,49 +98,46 @@ CHECK_C_SOURCE_COMPILES("
...
 CHECK_C_SOURCE_COMPILES("
-#if defined(_MSC_VER)
-#include <intrin.h>
-#else
-#include <x86intrin.h>
-#endif
-int main() {
-  __m256i r = _mm256_abs_epi32(_mm256_set1_epi32(1));
-}" COMPILER_SUPPORTS_AVX2)
+  #if defined(_MSC_VER)
```
fpetrogalli-arm: Same, revert

Configure.cmake
```diff
@@ -78,49 +98,46 @@ CHECK_C_SOURCE_COMPILES("
...
-#include <x86intrin.h>
-#endif
-void f(void *p) {
-  _mm512_loadu_si512(p);
-}
-int main() {
-  __m512i a = _mm512_set1_epi32(1);
-  __m512i r = _mm512_andnot_si512(a, a);
-}" COMPILER_SUPPORTS_AVX512F)
+  #if defined(_MSC_VER)
```
fpetrogalli-arm: Please revert the C code. The refactoring does not include cosmetic changes in the code.

src/libm-tester/CMakeLists.txt
```diff
@@ -7,7 +7,8 @@ include_directories(${CMAKE_BINARY_DIR}/src/libm/include) # rename headers
 
 set(LIB_SLEEF sleef)
 find_library(LIB_MPFR mpfr)
-set(EXTRA_LIBS m
+set(EXTRA_LIBS
+  m
```
fpetrogalli-arm: Please revert, this is not part of the refactoring

Configure.cmake
fpetrogalli-arm: also indenting the code is not part of the refactoring. Please revert this chunk, here and in the other places I commented yesterday.

diaena: Sorry, I have kind of lost track of these changes and it would take significant time for me to revert these with big impact on changes I have made that are based on this patch :(.
I am aware of my mistake and I apologise for the unrelated changes. Could we please let these in this patch and I will make sure no more of this happens. Thanks.

fpetrogalli-arm: I understand. Fine by me.

shibatch:
Could you check if the AVX512F code compiles on your environment?

diaena:
@shibatch The code does not compile and I get the same errors with "-O3 -DNDEBUG" (which are automatically added by cmake after setting the build type to Release).
I'm using Apple LLVM version 8.1.0 (clang-802.0.42).

```
sleef/src/arch/helperavx512f.h:350:61:` error: argument to '__builtin_ia32_pslldi512_mask' must be a constant integer
static INLINE vint2 vsll_vi2_vi2_i(vint2 x, int c) { return _mm512_slli_epi32(x, c); }
                                                            ^~~~~~~~~~~~~~~~~~~~~~~
```

shibatch:
I don't know which version of LLVM it corresponds to.
Could you install gcc and try it again?

diaena:
Ok, so the AVX512 code compiles successfully with GCC@5.4.1 on Linux. 
However, I get a `error: instruction requires: AVX-512 ISA` on MacOSX with GCC@5.4.0.
I tried a few different gcc versions and flag combinations, but didn't get to the bottom of the problem. I suspect it's an issue with the interaction between the GNU toolchain and the LLVM assembler?

shibatch:
Yes, there is that problem with OSX.
But we need to somehow check the AVX512 code.
Don't you have access to a linux computer?

diaena:
Yes, I said in the first line of my reply that I tested that the code compiles on Linux with GCC@5.4.1.

diaena:
Do we ignore the failure on MacOSX?

shibatch:
Ah, okay.
Let's ignore the failure on MacOSX, since there is no way to compile SLEEF on it.

PR #59, opened 2017-08-21
Default to optimised build to enable full support.

diaena:
I think sleef should have "Release" as the default cmake build type because there is code that require the optimisation flags to be turned on.
I have wrote the code such that if a build type if specified, our default will not overwrite the user's preference.

### DIFF
--- a/Configure.cmake
+++ b/Configure.cmake
@@ -136,19 +136,13 @@ CHECK_C_SOURCE_COMPILES("
   #else
   #include <x86intrin.h>
   #endif
-  void f(void *p) {
-    _mm512_loadu_si512(p);
+  __m512 addConstant(__m512 arg) {
+    return _mm512_add_ps(arg, _mm512_set1_ps(1.f));
   }
   int main() {
     __m512i a = _mm512_set1_epi32(1);
     __m512i r = _mm512_andnot_si512(a, a); }"
   COMPILER_SUPPORTS_AVX512F)
-
- # Turn AVX512F off for MacOSX until the assembler issue is resolved
-if(APPLE)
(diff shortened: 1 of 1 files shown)


shibatch:
We can rewrite the corresponding part with macro in order to make SLEEF compilable without optimization, but I think the current implementation is okay.

shibatch:
You don't need to turn off COMPILER_SUPPORTS_AVX512F for MAC OSX, since it is correctly detected by CHECK_C_SOURCE_COMPILES.

diaena:
@shibatch For some reason, the check for AVX512F passes with gcc on MacOs.
See this build log: https://travis-ci.org/shibatch/sleef/builds/266880810
image

shibatch:
I still don't fully understand what the problem is.
If check for AVX512F passes with gcc on MacOs but it does not build, isn't it possible to make the check fail by adding the failing intrinsic functions to the check?

diaena:
I think this: https://travis-ci.org/shibatch/sleef/jobs/266837612 is an assembler failure.
I cannot figure out which intrinsic to add to make it fail. If you have a second to take a look and tell me I could very quickly add it.

shibatch:
It seems quite everything is failing. Is this actually a problem in cmake? 

Let's try the following code, and if it does not work, let's add the exceptional rule for Mac OSX.

https://gcc.gnu.org/ml/gcc-bugs/2016-08/msg00205.html

diaena:
It is not a cmake failure, It fails the same way outside cmake, I have compiled manually a small piece of code to test these things.

I've came across that link after and it says the same thing, that it's a llvm assembler failure. I am pushing the code anyway just to prove the point.

PR #60, opened 2017-08-23
This patch fixes a bug that was making the error of sinpi larget than…

shibatch:
… the specification on very rare occasions.


fpetrogalli-arm:
May I ask you to add a test for these specific values?

xoofx:
> Please squash the commits, and thank you for using the commit message format discussed in the mailing list.

As this PR seems to fix an important bug, could we merge it?

PR #63, opened 2017-08-31
[gnuabi] Enable "finite" names (Double Precision).

fpetrogalli-arm:
glibc/libm defines "finite" names for some of the C99 functions.


fpetrogalli-arm:
@shibatch , gentle ping.

fpetrogalli-arm:
OK, I'll rework the patch. I didn't use aliases because I wanted to make it generic for any compiler - I believe that aliases work only for gcc and clang. This is fine by me. Thank you for your feedback.

shibatch:
I thought the current code is acceptable if the generated code is exactly the same. So, I now checked if gcc generates the same code for the two functions in a similar situation. It generates an extra jmp instruction for the second function. It could worsen in more complex situations, so I think it is better to take a measure at this time.

fpetrogalli-arm:
The current code uses aliases, there should not be any performance change. This set of patches are only for double precision, I will add the single precision ones when you are happy. I have added a test that checks the presence of the finite symbols in `libsleefgnuabi`.

fpetrogalli-arm:
@shibatch, gentle ping.

shibatch:
What I was thinking is that availability of alias attribute is checked within cmake or by checking the OS type, and it switches between the #define version and the alias version. How do you think?

fpetrogalli-arm:
That's reasonable, but the cmake build system is still incomplete, I want to do it in the Makefile system. Here is is safe to assume that ENABLE_GNUABI implies the availability of the `alias` attribute.

shibatch:
I thought that the problem occurs with OSX?

PR #68, opened 2017-09-24
Various fixes to compile SLEEF on Windows with MSVC

xoofx:
So this is a continuation on the `cmake-transition` branch in order be able to compile SLEEF on Windows with MSVC.

I have tested the compilation on both Windows+MSVC and Ubuntu 14.04 GCC and it seems to work fine there at least (though can't guarantee the travis build here, as it looks more finicky...

I have simplified the generation of the `renameXXX.h` files and `sleef.h` header file, by removing the need for `cmake -E` external files and instead using proper `add_custom_command`

This will also enable generation of LLVM bitcode directly. I may push this as part of this PR later (as this is quite small)

cc: @diaena As you did most of the work on the cmake branch, let me know what do you think about these changes.

Thanks!


xoofx:
Ha... actually, the sleef.h is getting screwed... cmake weirdness with arguments that I'm gonna have to debug...

xoofx:
So I have fixed most of the issues that I couldn't see on my test machine.

There are 2 remaining issues:
- For all GCC builds, Travis seems to finish all the builds/tests correctly but only fails at the very end with a return code of `1`... 
- For Clang, the AVX512F was not working (some missing AVX512F symbols not found). Tried to disable it for testing, but there was other problems there... not sure what is going on...

shibatch:
We are seeing many problems when compiling AVX512F code with clang. Please add symbols that cause problems to the detection code.

BTW, do you have Intel compiler?

shibatch:
I prepared an AppVeyor page for sleef.
If you want to enable it, please add appveyor.yml.

https://ci.appveyor.com/project/shibatch/sleef

src/libm/CMakeLists.txt
```diff
@@ -175,7 +179,43 @@ else()
   )
 endif()
 
+# --------------------------------------------------------------------
+# Generate LLVM bitcode
+# --------------------------------------------------------------------
+if(CLANG_EXE_PATH AND SLEEF_ENABLE_LLVM_BITCODE)
+  set(SLEEP_LLVM_BITCODE_INCLUDES "")
```
shibatch: SLEEP should be SLEEF.

xoofx: oops sorry 😅

fpetrogalli-arm:
Hello @xoofx , thank you for your patch!

Do you mind splitting it in separated PRs? 

I would like to see at least the following subdivision:

1. fixing AVX512 problem on clang, which has the highest priority as any PR seems to be broken right now. I wouldn't disable AVX512, I would just add an additional intrinsic (the one hat is failing) in the C code that gets tested to enable AVX512, #ifdef guarded to test it only when the compiler is clang.
2. replace the build time runs to generate the header files with the add_custom_command invocation (good idea, thanks for taking care of this)
3. building on MSVC, possibly activating appveyor in the same patchset.
4. targeting bitcode generation

Sorry for asking this, but it will make more easier for us to review them.

Thanks again!

Francesco

src/libm/sleefqp.c
```diff
@@ -22,10 +22,6 @@
 #pragma fp_contract (off)
 #endif
 
-#if defined(__GNUC__)
```
shibatch: Could you explain why you removed this?

xoofx: because it was failing compiling as it is already define in misc.h. So redefinition occured...

shibatch: I see. You are very quick!

xoofx:
> Do you mind splitting it in separated PRs?

Ok gonna try

xoofx:
> fixing AVX512 problem on clang, which has the highest priority as any PR seems to be broken right now. I wouldn't disable AVX512, I would just add an additional intrinsic (the one hat is failing) in the C code that gets tested to enable AVX512, #ifdef guarded to test it only when the compiler is clang.

This was fixed later with a more correct compilation detection.

> Do you mind splitting it in separated PRs?

@fpetrogalli-arm  Could I merge 1) and 2) together? As there are too many fixes in 2) (the previous system was completely silently continuing if mkrename was failing for example). and I would find cumbersome to separate...

fpetrogalli-arm:
@xoofx , yes, you are right, 1) and 2) make sense together. Thank you.

xoofx:
@fpetrogalli-arm great thanks! I will open a PR for 1) and 2), and once it is merged, I will open a PR for the LLVM bitcode part. Do you want me to squash the commits?

fpetrogalli-arm:
@xoofx, yes, please squash the commits. The main reason is that we would like to keep PRs related to "commit size" changes, as we believe it will make them easier to review.

Also, I kindly ask you to follow the commit message convention we established here: https://sourceforge.net/p/sleef/mailman/message/36008801/

Great stuff, thanks!

xoofx:
Great, closing this issue. Will reopen a PR for 1) and 2)

PR #70, opened 2017-09-25
[cmake] Add support for MSVC

xoofx:
This is a followup of the previous PR #69 in order to bring support for MSVC. 

I will use this PR to get appveyor working.

PR #73, opened 2017-09-26
[cmake] Add support for generating LLVM bitcode files

xoofx:
This is a continuation on top of #69, #70 and #72 

Currently, this can be only activated by passing `-DSLEEF_ENABLE_LLVM_BITCODE=1 ` to cmake.
It outputs the LLVM bitcode ll files to lib for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,8 @@ set(TARGET_IUTAVX2 "iutavx2")
 set(TARGET_IUTAVX2128 "iutavx2128")
 set(TARGET_IUTAVX512F "iutavx512f")
 set(TARGET_IUTADVSIMD "iutadvsimd")
+# The target to generate LLVM bitcode only, available when SLEEF_ENABLE_LLVM_BITCODE is passed to cmake
+set(TARGET_LLVM_BITCODE "llvm-bitcode")
 # Runs the test suite
 # Depends on targets: iut, tester
 # Defined in src/libm-tester/CMakeLists.txt via command add_custom_target
--- a/Configure.cmake
+++ b/Configure.cmake
@@ -98,24 +98,41 @@ command_arguments(RENAME_PARAMS_GNUABI_ADVSIMD advsimd n 2 4 float64x2_t float32
 
 # COMPILER DETECTION
 
+# Detect CLANG executable path (on both Windows and Linux/OSX)
+if(NOT CLANG_EXE_PATH)
+  # If the current compiler used by CMAKE is already clang, use this one directly
+  if(CMAKE_C_COMPILER MATCHES "clang")
+    set(CLANG_EXE_PATH ${CMAKE_C_COMPILER})
+  else()
+    # Else we may find clang on the path?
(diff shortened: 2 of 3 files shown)


xoofx:
So for this 1st iteration on generation of LLVM bitcode files:

1) The option is only activated if you pass `-DSLEEF_ENABLE_LLVM_BITCODE=1` to cmake
2) The LLVM LL files are generated to the lib folder

I have just a few stuffs to add to this branch for now:
- A small part in the doc cmake build that explain the option
- Activate the LLVM bitcode gen on the appveyor build for now

@shibatch and @fpetrogalli-arm, do you have any feedback on the above points?

shibatch:
I don't know the true reason that you need generation of LLVM bitcode, but if it is because Apple requires it, then isn't is easier to add options to generate fat binaries?

xoofx:
> I don't know the true reason that you need generation of LLVM bitcode, but if it is because Apple requires it, then isn't is easier to add options to generate fat binaries?

The reason is that I'm using SLEEF functions to inject them into a custom JIT/compiler before LLVM optimizations on the IR are happening (this is where LLVM performs most of its optimizations - constant folding, inlining, auto-vectorization...). Typically, if you want to do LTO (link time optimizations) with LLVM, you need to compile the code to LLVM bitcode instead of obj files.

shibatch:
I haven't tried these bitcode things, but did you consider making all bitcode files into a single archive? Then I think the option for bitcode generation can be turned on by default.

https://stackoverflow.com/questions/17459800/creating-and-using-llvm-bitcode-libraries

xoofx:
> I haven't tried these bitcode things, but did you consider making all bitcode files into a single archive? Then I think the option for bitcode generation can be turned on by default.

I don't have a particular need for merging them into a single archive but If I had, I would do it with llvm-link indeed.

> Then I think the option for bitcode generation can be turned on by default.

Hm, not sure to follow you here... Turn on by default for/when?

shibatch:
It is basically another form of library, right?
We can just simply build everything by default.
I am now also reading the web to check if the same thing is possible with gcc.

xoofx:
> It is basically another form of library, right?
> We can just simply build everything by default.

Actually, the LLVM bitcode in textual format is not meant to be backward compatible. But the binary format has been compatible since LLVM 3.0 (they don't say that it will be compatible forever but so far it has been the case). But not sure for now it is worth doing it. For example, the `llvm-link` binary is not part of the LLVM/CLang redistribuable on Windows so it would complicate our build story here. 
We can keep this internal for now as it is quite a very specific and not the most common usage of SLEEF. If there are other folks looking for this, we could think about how to do it properly.

shibatch:
Then Apple's attempt of requiring developers to embedding bitcode into applications is not very useful.

Anyway, please go ahead. This is not what we should care so much.

xoofx:
> Then Apple's attempt of requiring developers to embedding bitcode into applications is not very useful.

I don't know well the Apple story here, but their move to LLVM bitcode makes perfectly sense for an AppStore with multiple target platforms.

Anyway, thanks, I will merge this as-is for now. We can discuss about new opportunities with this later.

PR #77, opened 2017-10-02
[cmake] Fix libm dependency

fpetrogalli-arm:
On some Unix systems the libm dependency of libsleef and
libsleefgnuabi needs to be made explicit to avoid runtime failures in
the tests reporting "missing symbol: sqrt".

### DIFF
--- a/Configure.cmake
+++ b/Configure.cmake
@@ -4,6 +4,7 @@ include(CheckTypeSize)
 
 # Check that the library LIB_MPFR is available
 find_library(LIB_MPFR mpfr)
+find_library(LIBM m)
 
 # The library currently supports the following SIMD architectures
 set(SLEEF_SUPPORTED_EXTENSIONS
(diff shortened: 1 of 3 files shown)


shibatch:
Is it better to call __builtin_sqrt instead of sqrt?

fpetrogalli-arm:
@shibatch No, I believe that `__builtin_sqrt` is compiler dependent. Let's leave it as it is.

shibatch:
But then it does not allow using sleef as a substitute of libm.
__builtin_sqrt is compiler dependent, but we can check if those functions are available.

fpetrogalli-arm:
@shibatch I see your point. Let's make it an issue about removing completely libm dependency. Not for this patch.

@xoofx I added a tiny cosmetic change - mind taking a look at it again?

xoofx:
> @xoofx I added a tiny cosmetic change - mind taking a look at it again?

Actually yes, that's more accurate.

PR #78, opened 2017-10-02
[work-in-progress] Add glibc finite symbols in cmake

fpetrogalli-arm:
The two initial commits have been taken from master branch.

@xoofx , I need some help here. Do you have any idea how I could reproduce in cmake those "nm + grep" tests I have added in the Makefile of src/libm-tester?


xoofx:
I'm not sure exactly what the original makefile is doing (like checking if the symbol exist in the generated lib and output an error if not?)

Otherwise, I guess you can do the same using [file](https://cmake.org/cmake/help/v3.9/command/file.html?highlight=file) command, something in the line of:

```
    file(STRINGS ../../lib/libsleefgnuabi.$(DLLSUFFIX)  MYVAR_CHECK  REGEX __exp_finite)
```

fpetrogalli-arm:
The check is just making sure that the "finite" names introduced with the patch are available. The command "nm" is listing to stdout the symbols in the object file. All I am doing is a grep of that output. It is not very accurate, as it should be done for each of the SIMD extensions, but I wanted to keep it simple. I suspect I will have to write some C code for those symbols and interface them to the library just to make sure that are available.

Your suggestion with `file` could be a starting point for the original makefile test if I could get the path of the shared object `libsleefgnuabi`.

xoofx:
>  I suspect I will have to write some C code for those symbols and interface them to the library just to make sure that are available.

Yep, that would be safer and actually even less cumbersome than doing this with cmake commands...

fpetrogalli-arm:
Then, it is going to be easier if we turn on the ctest `enable_testing()/add_test()` mechanism and replace the `test-libm` target with the default one (`test`) provided by ctest.

I tried yesterday to do something in that direction in a separate patch but I couldn't get it right on the ${TARGET_TESTER} dependencies. I'll give it another go - I might submit a failing patch asking for help if I fail again.

xoofx:
>  I'll give it another go - I might submit a failing patch asking for help if I fail again.

Sure, I can have a look.

fpetrogalli-arm:
@xoofx , the build failure under OSX seems to be a problem with the homebrew configuration of the OSX runs on travis-ci: https://travis-ci.org/shibatch/sleef/jobs/285218079#L1577

Do you have any idea how to fix this?

xoofx:
Never used travis before.. there is this `MATRIX_EVAL` that is failing it seems, pulled-out from a config somewhere else but does not contain anything?

src/libm-tester/iutsimdmain.c
```diff
@@ -6,44 +6,9 @@
...
-#include <setjmp.h>
-
-static jmp_buf sigjmp;
+#include "check_feature.h"
 
 int do_test(int argc, char **argv);
-void check_featureDP();
-void check_featureSP();
-
-static void sighandler(int signum) {
```
fpetrogalli-arm: @shibatch Do you have any idea why by moving this SIGILL handler in a separate module the unsupported FMA4  iut tests started failing? See https://travis-ci.org/shibatch/sleef/jobs/286683373#L741

fpetrogalli-arm:
@xoofx gentle ping.

fpetrogalli-arm:
@shibatch I have added you for the review of the changes in `finite_name_testing.c`. Let me know if this is OK for you. As I said, I will add all the remaining "finite" functions in a separate pull request.

fpetrogalli-arm:
@xoofx , @shibatch : gentle ping

fpetrogalli-arm:
The test is just making sure that the code links against the "finite" symbols. I am calling the functions to make sure they don't get removed by the compiler if not used. I could probably just run the compiler with `-c` and just build an object files with a set of function using the `finite` symbols, but the actual executable is easier to build as a target to include in CTests.

PR #80, opened 2017-10-03
[cmake] Cmake uses ctests

fpetrogalli-arm:
The old `test-libm` custom target have been replaced by the facilities provided with ctests.


fpetrogalli-arm:
@xoofx , gentle ping. :)

PR #81, opened 2017-10-09
[fmod] Improve fmod

shibatch:
This patch is tested to work with AARCH32.


xoofx:
@fpetrogalli-arm can't we just merge master to `cmake-transition`? (I did this previously)

fpetrogalli-arm:
@xoofx Yes, of course. I was just trying to get only the changes for C code, not the makefile ones. But yea, it is absolutely fine to merge master into `cmake-transition`.

PR #83, opened 2017-10-09
[revert] This is just a test for CI.

fpetrogalli-arm:
Is this a real falure of SLEEF or a CI configuration poblem?

    https://travis-ci.org/shibatch/sleef/jobs/285218079

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ compiler:
 matrix:
   include:
     - os: osx
-      osx_image: xcode8
+      osx_image: xcode9.1
       compiler: gcc
       env:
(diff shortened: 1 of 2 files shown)


fpetrogalli-arm:
@xoofx @shibatch So, yes, CI is broken on OSX due to the ruby interpreter being out of date with what homebrew expects as minimal version: https://travis-ci.org/shibatch/sleef/jobs/285452667#L1577

shibatch:
Can't you use MacPorts instead?

fpetrogalli-arm:
@shibatch I'd rather not use MacPorts because `homebrew` seems to be "natively" supported by Travis CI.

fpetrogalli-arm:
I have self-approved this one because we need CI back again asap.

PR #87, opened 2017-10-12
[travis] Run tests in parallel.

fpetrogalli-arm:
This patch runs the build and tests in parallel, with two parallel
runs which are available according to [1].

It also removes the verbose output.

[1] https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,8 @@ before_install:
 before_script:
   - mkdir sleef.build
   - cd sleef.build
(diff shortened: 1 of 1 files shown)


shibatch:
Usually the number of parallel jobs should be somewhat higher than the core count, since there is a large amount of I/O in compilation. But it depends on other jobs running on the same computer in this case.

I think running only 2 jobs in parallel is okay since it would be always faster than 1 job.

fpetrogalli-arm:
I am going to merge this one because it is a non controversial change that will make further development easier by speeding up the testing process.

PR #89, opened 2017-10-14
Various cmake fixes

fpetrogalli-arm:
This patch-set implements fixes for the cmake build system, on aarch64, and also reactivate scalar testing for all architectures.

### DIFF
--- a/Configure.cmake
+++ b/Configure.cmake
@@ -70,7 +70,7 @@ elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
   set(SLEEF_HEADER_LIST
     ADVSIMD
   )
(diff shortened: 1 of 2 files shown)


Configure.cmake
```diff
@@ -70,7 +70,7 @@ elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
   set(SLEEF_HEADER_LIST
     ADVSIMD
   )
-  set(HEADER_PARAMS_ADVSIMD    "2 4 float64x2_t float32x4_t int32x2_t int32x4_t __ARM_NEON advsimd")
+  set(HEADER_PARAMS_ADVSIMD    2 4 float64x2_t float32x4_t int32x2_t int32x4_t __ARM_NEON advsimd)
```
xoofx: Ideally, you should use the `command_arguments` as for the other PARAMS, as this function make sure that everything is properly escaped (and will work on unix/windows).

fpetrogalli-arm: Thanks!

PR #92, opened 2017-10-17
This patch improves computation speed of asin and acos.

shibatch:
This patch reduces the length of asin kernel and speed-up computation speed of asin and acos functions.
The ratio of computation time between the previous xasin_u1 and the new one is 126%.


shibatch:
It's ratio between old and new execution time, and I stopped just indicating x% speedup. Isn't it simple enough?

https://en.wikipedia.org/wiki/Ratio

PR #96, opened 2017-10-19
[Performance] This patch improves computation speed of log1p.

shibatch:
Instead of calculating log1p with the generic log subroutine, the new functions calculate it using a new algorithm, which is a modification of generic log.
The ratio of computation time between the old and new log1p is 1.3.
This patch also includes some cosmetic changes and fixes of typos (ENABLE_FMA_DP was wrongly used in sleefsimdsp.c, and now replaced with ENABLE_FMA_SP).

### DIFF
--- a/src/libm/sleefdp.c
+++ b/src/libm/sleefdp.c
@@ -1777,16 +1777,46 @@ EXPORT CONST double xlog10(double d) {
   return r;
 }
 
-EXPORT CONST double xlog1p(double a) {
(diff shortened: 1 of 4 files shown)


fpetrogalli-arm:
May I ask you to cherry pick it or merge it into `cmake-transition`?

shibatch:
I cherry picked the last two branches for performance improvement into cmake-transition.

Issue #9, opened 2017-03-13
Micro benchmarks


fpetrogalli:
Use the google benchmark library [1] to write micro-benchmarks for the vector math routines.

Each benchmark should be invoking the function with random values in the input range.

[1] https://github.com/google/benchmark

shibatch:
I am now considering to overhaul the benchmarking tools.
My plan is to automate graphing with gnuplot first, and then introduce Google's framework.
In order to process data, I am planning to use java, since I am used to it.

fpetrogalli-arm:
I think we should do this after the transition from Makefile to cmake, as all the components will be easier to integrate.

I think that it would be better to implement this in the following order:

1. write the micro benchmarks with the google benchmark framework;
2. produce the graphic visualization.

I am happy for you to use Java+gnuplot, but I think that  it would be better to use python+mathplotlib, as it might be easier to maintain and run. It will require less configuration than setting up a Java VM, and I think it will be easier to deploy on Travis.

shibatch:
Do we need to deploy it on Travis?
You know, the benchmark results cannot be trusted since Travis is on a cloud.
The problem with python is that I need time to learn those things.
I have lots of experience with Java, and some experience with gnuplot, so it would be much easier for me.

fpetrogalli-arm:
Then, Java + gnuplot is. But, I still think you should first use the microbenchamrk system of google and then plot the data, because the output of Google benchmark might be different from what you expect in the plotting scripts. You might need to rework the scripts if that is going to be the case. Also, Google benchmark has some report facilities (see [1] for example), which might be used instead of the plotting tool.

I recommend you to leave the scripts for plotting as the last (optional) part of this work.

[1] http://1.bp.blogspot.com/-wk7hsdYodo8/UtS75FZag6I/AAAAAAAAAks/cQFiCXPbtwk/s1600/image00.png

shibatch:
Could you explain a little bit about how you are planning to use the output data?
I have been always think this work as a part of writing our paper.
From that perspective, drawing graph is essential.

fpetrogalli-arm:
Getting numbers is very important. Using google framework should guarantee that we have reliable evaluations. If the aim of setting up this is only to get the numbers in the paper, I think we don't need to store any script in here for that. 

My goal here was to make sure that the numbers we were getting where reliable, and I believe that those we could get with google micro benchmarks framework have such property. They would be reliable not only for us, but also for people reading these numbers are they are produce via a standard tool.

[...]

PR #98, opened 2017-10-27
[Benchmark] New benchmarking tool with gnuplot

shibatch:
This patch replaces the old benchmarking tool with a new tool with gnuplot.

With this new tool, the graph showing execution time can be automatically drawn.
It is now easy to see the difference between different library versions or compilers.
Some results are shown at http://shibatch.github.io/

This patch implements the tool described in https://github.com/shibatch/sleef/issues/9.
I am planning to add another benchmarking tool based on the Google microbenchmarking framework.


fpetrogalli-arm:
Are these benchmarks specific to Intel? In #9 the idea was to write micro-benchmark that could work on any architecture, using the googletest tool as agreed, and using the same mechanisms of the helper header files that is used to write target independent tests.

Please make sure they build (not run!) on the Travis-CI machines. I might have missed it, but I can't see the build output on CI.

shibatch:
At this point, the benchmarking tool is specific to Intel. I am planning to expose all functions with names without vector extension name, and after making that change, it will support ARM.

I was also looking for a way to run the benchmarking tool on travis, but how do you think about this? It requires an Amazon S3 account, but a free license expires after the first 12 months.

fpetrogalli-arm:
@shibatch , I understand that these are the benchmarks that are used to produce the figures for release 3.2, as explained in #147. Please rebase against master and build them with cmake. Don't merge master, otherwise it will be cumbersome as all the changes since the PR will be shown. Also, I would like to break the contents in two separate folders:
1. put the benchmark code in src/libm-benchmarks
2. put the java/gnuplot tools in src/libm-benchmarks/plot-scripts

shibatch:
Rebased. The new PR is https://github.com/shibatch/sleef/pull/168.

PR #103, opened 2017-11-16
Merge cmake transition into master

fpetrogalli-arm:
This branch merges the `cmake-transition` branch into master.

Now we have two different build system coexisting  in master. The situation is as follows:

# Libdft
1. build with Makefile
2. tested via Makefile with target `make test-dft`.

# Libsleef and libsleefgnuabi
1. both built with Makefile _AND_ cmake
2. tested only via the cmake configuration

The Makefile is suddenly being used in CI for a variate of OS/$CC configurations which were previously not tested on Travis, so I expect things to break. I will fix this along the way in this patch.

The actual merge commit from which the fixes are being added is: https://github.com/shibatch/sleef/commit/c1d920c7c29ef7f002a2ed8c5fa6d6fcf6d8636c


fpetrogalli-arm:
The merging point is `merge-point-of-cmake-transition-into-master`. This branch consists only of the set of commits on top of the merging point to fix the build failures on Travis-CI.

shibatch:
Isn't it better to change the name of branch?

fpetrogalli-arm:
I am not sure I understand. The branch is called `merge-cmake-transition-into-master`, and I want you to review it against `merge-point-of-cmake-transition-into-master`. Once you accept it, I will merge `merge-point-of-cmake-transition-into-master` into `master`. I did it that way because the pull request of `cmake-transition` into master was showing all the changes in `cmake-transition`, which was pointless because those changes had already been reviewed. At the end of this process I'll delete all the `cmake-transition`-related branch an everything will just be in master.

shibatch:
I meant the name "Merge cmake transition into master". Isn't it saved on the git?

fpetrogalli-arm:
Oh I see - I plan to squash the commits in a single commit, I will change the head line in something like "[build] Add cmake build system".

xoofx:
> Oh I see - I plan to squash the commits in a single commit, I will change the head line in something like "[build] Add cmake build system".

btw, when squashing, could you squash commits by consecutive authors instead of squashing everything into a single commit, it doesn't help to track original code, who can help actually on a particular part of the build...etc. But when merging a development branch like this, I would keep everything as it is, otherwise you are erasing lots of information (when stuff happened, who did it...etc.)

fpetrogalli-arm:
@xoofx I have squashed only my commits to fix the merge. The history of `cmake-transition` is kept in https://github.com/shibatch/sleef/pull/104. :)

xoofx:
good, perfect then.

PR #105, opened 2017-11-20
Use sleef.lib on Windows. Export functions with extern "C"

xoofx:
This is a fix for Windows for importing correctly `sleef.lib` and exporting functions using extern "C" so that they are correctly exported in the DLL (without C++ mangling names)

I have updated the way the headers are concatenated with now a `sleeflibm_header.h.org` and `sleeflibm_footer.h.org`

### DIFF
--- a/src/libm/CMakeLists.txt
+++ b/src/libm/CMakeLists.txt
@@ -6,7 +6,8 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 # sleef.h
 # --------------------------------------------------------------------
 # File generated for the headers
-set(SLEEF_ORG_HEADER ${CMAKE_CURRENT_SOURCE_DIR}/sleeflibm.h.org)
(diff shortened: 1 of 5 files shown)


fpetrogalli-arm:
I am not keen in developing this in Makefiles. We would like to make the cmake system the official build system and deprecate the Makefile one. @shibatch , are you happy for me to ask @xoofx to remove the Makefile changes in this patch? In fact, we are not even testing on appveyor the windows builds with Makefiles.

xoofx:
> I am not keen in developing this in Makefiles.

Yeah, but that's mainly to have them still working, otherwise legacy Makefiles would not work anymore with this PR.

fpetrogalli-arm:
Oh I see! Makes sense. Thanks

fpetrogalli-arm:
> Merged #105.

Thanks!

PR #106, opened 2017-11-24
* make xfmaxf(), xfminf(), xfmax() and xfmin() aware of ISAs with min…

blu:
…max-number capabilities via macro SLEEF_*_MINMAXNUM_AVAILABLE

* add minmax-number capabilities to arch AArch64

### DIFF
--- a/src/libm/sleefsimdsp.c
+++ b/src/libm/sleefsimdsp.c
@@ -1678,11 +1678,19 @@ EXPORT CONST vfloat xfabsf(vfloat x) { return vabs_vf_vf(x); }
 EXPORT CONST vfloat xcopysignf(vfloat x, vfloat y) { return vcopysign_vf_vf_vf(x, y); }
 
 EXPORT CONST vfloat xfmaxf(vfloat x, vfloat y) {
+#if SLEEF_SINGLE_MINMAXNUM_AVAILABLE
+  return vmaxnum_vf_vf_vf(x, y);
+#else
   return vsel_vf_vo_vf_vf(visnan_vo_vf(y), x, vsel_vf_vo_vf_vf(vgt_vo_vf_vf(x, y), x, y));
+#endif
 }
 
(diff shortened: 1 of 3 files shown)


shibatch:
Hi Martin,

Thank you for your contribution.
Could you please introduce yourself a little bit?

shibatch:
That's great!
I found the corresponding pull request.

https://github.com/pocl/pocl/pull/539

blu:
Yes, from what I can tell, [#539 ](https://github.com/pocl/pocl/pull/539) was the original SLEEF-into-POCL pull; my modification came later in [#564](
https://github.com/pocl/pocl/pull/564).

src/libm/sleefsimdsp.c
```diff
@@ -1678,11 +1678,19 @@ EXPORT CONST vfloat xfabsf(vfloat x) { return vabs_vf_vf(x); }
 EXPORT CONST vfloat xcopysignf(vfloat x, vfloat y) { return vcopysign_vf_vf_vf(x, y); }
 
 EXPORT CONST vfloat xfmaxf(vfloat x, vfloat y) {
+#if SLEEF_SINGLE_MINMAXNUM_AVAILABLE
```
fpetrogalli-arm: Hello @blu , thank you for working on this. This is a good optimization.
My preference though would be to avoid the preprocessor code in the `sleefsimd[s|d]p.c` and just define the `vmaxnum_vf_vf_vf` and `vminnum_vf_vf_vf` (and their double precision counterparts) in the helper files.

To avoid duplicating the same definition in all the helper files that do not have the underlying direct instruction, would be to write the generic methods in `df.h` and `dd.h`, `#ifndef` guarded with the macros you created, and define those macros in the helper files that have such instructions (AdvSIMD at the moment).

This should minimize the amount of code written for this optimization, especially if we find other places in the library where we can use it.

fpetrogalli-arm:
@blu , can we drop this pull request now that #109  is merged?

PR #109, opened 2017-11-26
* add single- and double-precision vmaxnum()/vminnum() adhering to IE…

blu:
…EE-754-2008 maxNum()/minNum() semantics

* make xfmaxf(), xfminf(), xfmax() and xfmin() use vmaxnum() and vminnum()
* add vminnum()/vmaxnum() capabilities to arch SSE2, SSE3, SSE4.1, AVX, AVX2 and AArch64

### DIFF
--- a/src/arch/helperadvsimd.h
+++ b/src/arch/helperadvsimd.h
@@ -28,6 +28,16 @@
 
 #define ISANAME "AArch64 AdvSIMD"
 
+#ifdef SLEEF_SINGLE_MINMAXNUM_AVAILABLE
+#error prior definition of SLEEF_SINGLE_MINMAXNUM_AVAILABLE
+#endif
+#define SLEEF_SINGLE_MINMAXNUM_AVAILABLE 1
+
+#ifdef SLEEF_DOUBLE_MINMAXNUM_AVAILABLE
+#error prior definition of SLEEF_DOUBLE_MINMAXNUM_AVAILABLE
+#endif
+#define SLEEF_DOUBLE_MINMAXNUM_AVAILABLE 1
+
 // Mask definition
 typedef uint32x4_t vmask;
 typedef uint32x4_t vopmask;
@@ -149,6 +159,14 @@ static INLINE vfloat vmax_vf_vf_vf(vfloat x, vfloat y) {
 }
 static INLINE vfloat vmin_vf_vf_vf(vfloat x, vfloat y) {
   return vminq_f32(x, y);
+}
+
+// max number, min number
+static INLINE vfloat vmaxnum_vf_vf_vf(vfloat x, vfloat y) {
+  return vmaxnmq_f32(x, y);
+}
+static INLINE vfloat vminnum_vf_vf_vf(vfloat x, vfloat y) {
+  return vminnmq_f32(x, y);
 }
 
 // Comparisons
@@ -288,6 +306,14 @@ static INLINE vdouble vmax_vd_vd_vd(vdouble x, vdouble y) {
 }
 static INLINE vdouble vmin_vd_vd_vd(vdouble x, vdouble y) {
   return vminq_f64(x, y);
+}
+
+// max number, min number
+static INLINE vdouble vmaxnum_vd_vd_vd(vdouble x, vdouble y) {
+  return vmaxnmq_f64(x, y);
+}
+static INLINE vdouble vminnum_vd_vd_vd(vdouble x, vdouble y) {
+  return vminnmq_f64(x, y);
 }
 
 // Multiply accumulate: z = z + x * y
--- a/src/arch/helperavx.h
+++ b/src/arch/helperavx.h
@@ -29,6 +29,16 @@
 
 #define FULL_FP_ROUNDING
 
+#ifdef SLEEF_SINGLE_MINMAXNUM_AVAILABLE
+#error prior definition of SLEEF_SINGLE_MINMAXNUM_AVAILABLE
+#endif
+#define SLEEF_SINGLE_MINMAXNUM_AVAILABLE 1
+
+#ifdef SLEEF_DOUBLE_MINMAXNUM_AVAILABLE
+#error prior definition of SLEEF_DOUBLE_MINMAXNUM_AVAILABLE
+#endif
+#define SLEEF_DOUBLE_MINMAXNUM_AVAILABLE 1
+
 #if defined(_MSC_VER)
 #include <intrin.h>
 #else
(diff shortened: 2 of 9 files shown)


blu:
This is a redux of PR  #106, implementing the changes requested by @fpetrogalli-arm

blu:
Fix incoming..

src/arch/helperavx.h
```diff
@@ -29,6 +29,16 @@
 
 #define FULL_FP_ROUNDING
 
+#ifdef SLEEF_SINGLE_MINMAXNUM_AVAILABLE
```
fpetrogalli-arm: I must have been very unclear in my requests in #106 ! :)

Given that you have added the generic implementation of `v[max|min]num_vd_vd_vd` in ``dd.h``, guarded with an `#ifndef`, there is no need to provide it via the helper header files. You should just be able to revert the `avx*` header file changes you have introduced in this patch, so that the generic implementation in `dd.h` would be picked up. A similar reasoning holds for the single precision version of the functions.

So please remove the macros , and the generic implementations from the `axv*` header files.

src/arch/helperadvsimd.h
```diff
@@ -28,6 +28,16 @@
 
 #define ISANAME "AArch64 AdvSIMD"
 
+#ifdef SLEEF_SINGLE_MINMAXNUM_AVAILABLE
+#undef SLEEF_SINGLE_MINMAXNUM_AVAILABLE
```
fpetrogalli-arm: This `#ifdef`-guarded `#undef` is not what we want here. If for any reason the  `SLEEF_SINGLE_MINMAXNUM_AVAILABLE` macro is already defined when the C preprocessor reaches this point, we _do want_ to have an error here as it means that there is something wrong in the code, as the `SLEEF_SINGLE_MINMAXNUM_AVAILABLE` macro should be defined only once in all the compilation.
The preprocessor fires a warning in case of duplicate macro, but I think that an `#error` preprocessor invocation would be more appropriate.

blu:
@fpetrogalli-arm, change's been updated as per your review

src/libm/df.h
```diff
@@ -464,3 +464,13 @@ static INLINE CONST vfloat2 dfsqrt_vf2_vf(vfloat d) {
...
   return dfscale_vf2_vf2_vf(dfmul_vf2_vf2_vf2(dfadd2_vf2_vf_vf2(d, dfmul_vf2_vf_vf(t, t)), dfrec_vf2_vf(t)), vcast_vf_f(0.5f));
 }
+
+#ifndef SLEEF_SINGLE_MINMAXNUM_AVAILABLE
+static INLINE CONST vfloat vmaxnum_vf_vf_vf(vfloat x, vfloat y) {
+  return vmax_vf_vf_vf(vsel_vf_vo_vf_vf(visnan_vo_vf(x), y, x), vsel_vf_vo_vf_vf(visnan_vo_vf(y), x, y));
+}
+
+static INLINE CONST vfloat vminnum_vf_vf_vf(vfloat x, vfloat y) {
+  return vmin_vf_vf_vf(vsel_vf_vo_vf_vf(visnan_vo_vf(x), y, x), vsel_vf_vo_vf_vf(visnan_vo_vf(y), x, y));
```
fpetrogalli-arm: Apologies for raising another question. This is the original implementation of the code that is currently in `sleefsimdsp.c`. For the x86 helper files, you have changed such implementation with `return vsel_vf_vo_vf_vf(visnan_vo_vf(y), x, vmin_vf_vf_vf(x, y));`. The new one seems to be better, would it be possible to replace this old one with the new one? If the answer is yes, I would suggest you to make the generic implementation in `df.h` as `return vsel_vf_vo_vf_vf(visnan_vo_vf(y), x, vmin_vf_vf_vf(x, y));` and remove the code (`#ifdef`s and function definitions) you have added in the x86 helper files. This reasoning about vmin holds also for the vmax function and the double precision versions in `dd.h`.

blu: I'm not sure I follow. This is the original  (i.e. currently in master) implementation in `sleefsimdsp.c`:
```
EXPORT CONST vfloat xfminf(vfloat x, vfloat y) {
  return vsel_vf_vo_vf_vf(visnan_vo_vf(y), x, vsel_vf_vo_vf_vf(vgt_vo_vf_vf(y, x), x, y));
}
```
The original version is not really generic -- it's tuned to the semantics of the x86 `minps/maxps/vminps/vmaxps` instructions; that's why the new x86-specific versions of `vminnum/vmaxnum` are slightly-optimized versions of the original, with `vsel(vgt())` changed to `vmin()`.

The proposed in this PR generic versions of `vminnum/vmaxnum` found in `df.h/dd.h` are actually generic, i.e. agnostic of the peculiarities of the underlying min/max vector ops. The proposed implementation of generic `vminnum/vmaxnum` in `df.h/dd.h` and specialized versions in the respective ISA helpers is the most logical implementation of this functionality.

shibatch: The original implementation is optimized in terms of the number of operations. When min and max instructions are available, I think NAN handling is not required. So, it just needs one instruction, like the following.
```
EXPORT CONST vdouble xfmax(vdouble x, vdouble y) {
   return vmax_vd_vd_vd(x, y);
}
```

blu: @shibatch, aren't `xfmax()/xfmin()` supposed to implement IEEE754-2008 minNum/maxNum semantics? The way the original is implemented, it tries to adhere to minNum/maxNum assuming x86 unerlying semantics (which are no natively minNum/maxNum-compliant).

Here's quick example of how the original would behave:
```
x: num, y: nan
original_xfmin(x, y) -> x

x: nan, y: num
original_xfmin(x, y) -> y
```
In contrast to a x86 `vmin()`:
```
x: num, y: nan
x86_vmin(x, y) -> nan

x: nan, y: num
x86_vmin(x, y) -> y
```

shibatch: NAN handling should be implemented in those instructions.

http://www.felixcloutier.com/x86/MAXPD.html

blu: While x86 `minps/maxps/vminps/vmaxps` have some handling of NaNs, that handling is not IEEE754-2008/C99-compliant. To quote the semantics from that page your referred to:

>If only one value is a NaN (SNaN or QNaN) for this instruction, the second operand (source operand), either a NaN or a valid floating-point value, is written to the result.

In contrast to C99 (http://en.cppreference.com/w/c/numeric/math/fmin):
>If one of the two arguments is NaN, the value of the other argument is returned. Only if both arguments are NaN, NaN is returned

shibatch: Okay, I now understand. You are correct.

fpetrogalli-arm:
@shibatch , are you happy for me to merge this?

src/libm/df.h
shibatch: But the instructions in ARM is IEEE754-compliant.

http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0802b/CIHFCJCF.html

If we seriously trying to optimize this, it is better to make it choose the implementation according to IEEE754-compliance of the instructions.
I think this is a good opportunity to do this.

fpetrogalli-arm: @shibatch , are you saying that we should get completely rid of the old implementation and use the ones that @blu added in the x86 helper header files as generic implementation to be used in `[dd|df].h`?

This makes sense to me too.

blu: @fpetrogalli-arm, looking again at the new generic version, I don't think it would schedule better on any of the current popular uarchs than the original `xfmin()/xfmax()`. So I think the original `xfmin()/xfmax()` should become the generic version; the one issue with the new x86 version becoming the generic is that it won't work on ISAs with fmin/fmax semantics returning NaN when either of the args is NaN.

shibatch: In a situation like this, I usually write the exported functions (xfmin()/xfmax() in this case) in a generic way. Here, being generic means that the function returns the correct value regardless of the architecture. If I write an optimized version that works only for specific vector extensions, it is chosen with #if macro.

Basically, the exported functions (those functions beginning with x) should not be called by other implementation of functions. I also implement functions for internal usage, which do not handle non-numbers.

blu: @shibatch @fpetrogalli-arm Ok, give me a couple of hours (currently preoccupied) to show you what I mean.

PR #111, opened 2017-11-30
[cmake] Adding c make lists.txt for dft

shibatch:
This patch adds the files required to build DFT library with cmake.

There are a few things that are not working with this patch.
- Parallel build does not work with MSVC.
- It does not detect change of unroll0.org

I still don't understand the difference between source and target, so the way I write CMakeLists.txt might be strange. Please point out if there is things that can be improved.


Configure.cmake
```diff
@@ -36,7 +36,13 @@ function(command_arguments PROPNAME)
   set(${PROPNAME} ${quoted_args} PARENT_SCOPE)
 endfunction()
 
+# Detect the path of cmake executable
 
+find_program(CMAKE_EXE_PATH "cmake")
```
xoofx: There is already the `CMAKE_COMMAND` variable available

src/dft/CMakeLists.txt
```diff
@@ -0,0 +1,279 @@
...
+  string(CONCAT SG ${S} "_generated")                       # SG is "dispatchdp.h_generated"
+  add_dependencies(${G} ${SG})
+  set_target_properties(${G} PROPERTIES ${COMMON_TARGET_PROPERTIES})
+  list(GET LISTTYPEID ${T} ID)                              # ID is 1
+  target_compile_definitions(${G} PRIVATE BASETYPEID=${ID} ${COMMON_TARGET_DEFINITIONS})
+endforeach()
+
+# Copy unroll0.org to ${CMAKE_CURRENT_BINARY_DIR}
+
+add_custom_target(unroll0.org_source
```
xoofx: Better `add_custom_command` instead so that you can specify proper dependencies between input and output files

shibatch: This is where I am having trouble.
I thought that dependency between source ${CMAKE_CURRENT_SOURCE_DIR}/unroll0.org and target unroll0.org_source can be specified clearly with this.
I will check if add_custom_target works better tomorrow since I am away on business trip today.

.travis.yml
```diff
@@ -12,10 +12,10 @@ compiler:
 matrix:
   include:
     - os: osx
-      osx_image: xcode8
+      osx_image: xcode8.3
       compiler: gcc
       env:
-        - MATRIX_EVAL="brew install gcc@6 && CC=gcc-6" # install and use gcc
+        - MATRIX_EVAL="rm -f '/usr/local/include/c++' && brew install gcc@6 && CC=gcc-6" # install and use gcc
```
fpetrogalli-arm: This looks  unsafe. Why did you remove this?

Configure.cmake
```diff
@@ -36,7 +36,13 @@ function(command_arguments PROPNAME)
...
+
+# Detect if cmake is running on Travis
+string(COMPARE NOTEQUAL "" "$ENV{TRAVIS}" RUNNING_ON_TRAVIS)
```
fpetrogalli-arm: I think we should not hard-code Travis detection in the cmake file. Why do you need to detect if we are configuring the build on Travis?

Configure.cmake
```diff
@@ -136,6 +142,9 @@ if(CMAKE_C_COMPILER_ID MATCHES "(GNU|Clang)")
 
   # Warning flags.
   set(FLAGS_WALL "-Wall -Wno-unused -Wno-attributes")
+  if(CMAKE_C_COMPILER_ID MATCHES "GNU")
+    string(CONCAT FLAGS_WALL ${FLAGS_WALL} " -Wno-psabi")
```
fpetrogalli-arm: It would be good to have a comment saying why this option is needed to compile libdft.

Configure.cmake
```diff
@@ -156,10 +165,21 @@ if(LD_SIZE GREATER "9")
   set(COMPILER_SUPPORTS_LONG_DOUBLE 1)
 endif()
 
+# Bug workaround for gcc 4.x
```
fpetrogalli-arm: Is this something you need on your machine? Or is this something that is happening just on the version of Ubuntu that Travis runs? I'd prefer `COMPILER_SUPPORTS_LONG_DOUBLE`  to be detected by the script as it currently does. What is preventing this detection to work?

Configure.cmake
```diff
@@ -156,10 +165,21 @@ if(LD_SIZE GREATER "9")
...
+  if (CMAKE_C_COMPILER_VERSION VERSION_LESS 5.0)
+    set(COMPILER_SUPPORTS_LONG_DOUBLE 0)
+  endif()
+endif()
+
 CHECK_C_SOURCE_COMPILES("
   int main() { __float128 r = 1;
   }" COMPILER_SUPPORTS_FLOAT128)
 
+if (${RUNNING_ON_TRAVIS} AND CMAKE_C_COMPILER_ID MATCHES "Clang")
```
fpetrogalli-arm: I really think we shouldn't have any Travis-aware variable in the build system. What is the exact error that forces you to add this condition?

lib/Makefile.vc
```diff
@@ -2,22 +2,22 @@
 all : libsleef libsleefgnuabi libsleefdft
 
 .PHONY: libsleef
-libsleef : libsleef.dll
+libsleef : sleef.dll
```
fpetrogalli-arm: This changes in the Makefile build system are completely unexpected. Why do you need them? This patch is building libdft with cmake, it should ignore the Makefiles.

src/common/CMakeLists.txt
```diff
@@ -1,2 +1,21 @@
...
+  )
+
+# Target TARGET_LIBCOMMON_STATIC
+
 add_library(${TARGET_LIBCOMMON_STATIC} STATIC common.c)
+set_target_properties(${TARGET_LIBCOMMON_STATIC} PROPERTIES ${COMMON_TARGET_PROPERTIES})
+
+# Target common.o
+
+add_library(common_obj OBJECT common.c)
```
fpetrogalli-arm: Why do you need the object here? You should use the `${TARGET_LIBCOMMON_STATIC}` library built on line 10.

src/dft/CMakeLists.txt
xoofx: Note that SOURCES in add_custom_target doesn't express a dependency, but just notifies that the files should be visible to the IDE project. A custom target is always considered out-of-date and always built. That's why it is better to rely on `add_custom_command` (but you can then add a dependency between a custom target and a custom command)

src/dft/CMakeLists.txt
```diff
@@ -0,0 +1,279 @@
...
+      target_compile_definitions(${UG} PRIVATE ${COMMON_TARGET_DEFINITIONS} ${MACRODEF_${E}})
+      target_compile_options(${UG} PRIVATE ${CFLAGS_${E}})
+      add_dependencies(${UG} ${TARGET_HEADERS} ${UCG})
+    endforeach()
+  endforeach()
+endforeach()
+
+# Target libdft
+
+add_library(${TARGET_LIBDFT} SHARED $<TARGET_OBJECTS:dftcommon_obj> $<TARGET_OBJECTS:common_obj> $<TARGET_OBJECTS:arraymap_obj>)
```
fpetrogalli-arm: Instead of using `$<TARGET_OBJECTS:common_obj>`, couldn't you just link the original static library in the next line? Similarly, for `$<TARGET_OBJECTS:arraymap_obj>`: couldn't you have build it as a static library and linked it once in the next line? Am I missing something?

CMakeLists.txt
```diff
@@ -68,6 +68,10 @@ set(TARGET_MKDISP "mkdisp")
 # Defined in src/common/CMakeLists.txt via command add_library
 set(TARGET_LIBCOMMON_STATIC "common")
 
+# Generates object file (shared library) `libsleefdft`
+# Defined in src/dft/CMakeLists.txt via command add_library
+set(TARGET_LIBDFT "sleefdft_shared")
```
fpetrogalli-arm: I think the name `sleefdft` is enough, no need to append `_shared`. As you can see from https://travis-ci.org/shibatch/sleef/jobs/309529448#L1158, the resulting library has a .so extension that makes the fact that the library is a shared object explicit.

.travis.yml
shibatch: This is suggested by brew.
Without this, brew terminates with an error saying there is already a link /usr/local/include/c++.
Please let me know if there is another good way other than this.

lib/Makefile.vc
shibatch: Probably it is better to make this another pull request, but this is necessary since the cmake is already generating a library under the name sleef.dll. Without changing in this way, sleef.dll cannot be linked with Windows. See the change in sleeflibm.h.org.

Configure.cmake
shibatch: This is because linking with OpenMP fails on Travis although detection succeeds.

.travis.yml
fpetrogalli-arm: But this was not happening in previous builds. If you really cannot remove it, please add a comment explaining why it is needed, possibly with a link to the suggestion from brew, so that we know where it comes from! It looks ugly! :)

fpetrogalli-arm:
@shibatch , I just realized that you are asking to merge this pull request into `cmake-transition`. You should change the destination branch to `master`.

shibatch:
I want to merge the cmake-transition branch soon. After this pull request is approved, I see no big reason not to merge cmake-transition. I also want to start implementing https://github.com/shibatch/sleef/issues/47, which would require many changes in the CMakeLists.txts.

I believe that you want to keep cmake-transition because you are using it at your company. One possible way is to effectively merge this tree and cmake-transition tree into master with this pull request, and keep cmake-transition as it is now. Does this work for you?

xoofx:
@shibatch  the `cmake-transition` branch has been already merged to master

shibatch:
Ah, okay. :)

Configure.cmake
shibatch: This is now replaced with a detection code.

src/common/CMakeLists.txt
shibatch: What I want to do is to add those object files into a shared library. If they are object files, it can be done with $<TARGET_OBJECTS:objlib>, but how can I do this with a static library?

https://stackoverflow.com/questions/11429055/cmake-how-create-a-single-shared-library-from-all-static-libraries-of-subprojec

src/libm/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ set(SLEEF_ORG_HEADER ${CMAKE_CURRENT_SOURCE_DIR}/sleeflibm.h.org)
 set(SLEEF_INCLUDE_HEADER ${CMAKE_BINARY_DIR}/include/sleef.h)
 
 set(SLEEF_HEADER_COMMANDS "")
-list(APPEND SLEEF_HEADER_COMMANDS COMMAND cmake -E copy ${SLEEF_ORG_HEADER} ${SLEEF_INCLUDE_HEADER})
+list(APPEND SLEEF_HEADER_COMMANDS COMMAND ${CMAKE_EXE_PATH} -E copy ${SLEEF_ORG_HEADER} ${SLEEF_INCLUDE_HEADER})
```
xoofx: It should be `CMAKE_COMMAND` right now?

xoofx:
@shibatch as suggested by @fpetrogalli-arm you can change the target branch to master for this PR  😉

shibatch:
Do I need to close this pull request once and make it again in order to merge to master?

xoofx:
> Do I need to close this pull request once and make it again in order to merge to master?

xoofx:
> Do I need to close this pull request once and make it again in order to merge to master?

yeah, not sure you can do this with plain GitHub.... Maybe better to rebase your work on top on master and PR again...

PR #113, opened 2017-12-03
[AVX512F] Fix misdetection of AVX512F

shibatch:
This patch should fix misdetection of AVX512F availability happening mainly on macOS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ language: c
 os: linux
 sudo: required
 dist: trusty
-compiler:
-  - gcc
-  - clang
 
 # Include osx jobs
 matrix:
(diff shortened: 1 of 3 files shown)


shibatch:
CLANG_FLAGS_ENABLE_${SIMD} are copied to FLAGS_ENABLE_${SIMD} at line 127 [1].
Probably we had better do some cleanup of the CMakefile.txt.
One reason of misdetection is that detection code cannot be compiled[2].

[1] https://github.com/shibatch/sleef/pull/113/files#diff-f54cc8a74b3509affca6175972035ecaL127
[2] https://github.com/shibatch/sleef/pull/113/files#diff-f54cc8a74b3509affca6175972035ecaL225

fpetrogalli-arm:
If I understand correctly, here you are trying to make sure that GCC compiles for AVX512F on MacOS. It seems that this is still not happening [1].

This is strange because the same code compiles in this example: https://godbolt.org/g/ccUfBL

[1]https://travis-ci.org/shibatch/sleef/jobs/310786354#L1729

shibatch:
My understanding is that a mac-specific problem is preventing AVX512F from compiling on MacOS. Since all the elements in the detection code are required to compile libm, there is no way to compile it if detection fails.

PR #119, opened 2017-12-05
[travis] Enable gcc-7 and clang-5.0 on linux.

fpetrogalli-arm:
This is me experimenting with travis configuration to see if I can compile AVX512F at least in the linux environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ language: c
 os: linux
 sudo: required
 dist: trusty
-compiler:
-  - gcc
-  - clang
 
 # Include osx jobs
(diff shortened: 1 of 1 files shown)


fpetrogalli-arm:
AVX512F built on Linux, both with GCC [1] and CLANG [2].

[1] https://travis-ci.org/shibatch/sleef/jobs/311819180#L715
[2]  https://travis-ci.org/shibatch/sleef/jobs/311819181#L633

@shibatch , I think you should merge this change in your branch  `Fix_misdetection_of_AVX512F_availability`, and then into `master`.

PR #124, opened 2017-12-10
[Bit identity] Introducing tester3

shibatch:
This tester checks if the returned values from two implementations are bit-identical. This test is quick since it does not need libmpfr.
This patch is a part of implemention of https://github.com/shibatch/sleef/issues/47

### DIFF
--- a/src/libm-tester/CMakeLists.txt
+++ b/src/libm-tester/CMakeLists.txt
@@ -104,6 +104,8 @@ set(TESTER3_0 AVX2128        __m128d d2 __m128 f4 AVX2    __m256d d4 __m256 f8)
 set(TESTER3_1 SSE2           __m128d d2 __m128 f4 AVX     __m256d d4 __m256 f8)
 set(TESTER3_2 SSE4           __m128d d2 __m128 f4 AVX     __m256d d4 __m256 f8)
 
+add_library(tester3main_obj OBJECT tester3main.c)
+
 foreach(N 0 1 2)
   list(GET TESTER3_${N} 0 EXTX)                                    # EXTX is AVX2_SCALAR, for example
   list(GET TESTER3_${N} 1 DPTYPEX)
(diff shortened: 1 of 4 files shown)


src/libm-tester/tester3main.c
```diff
@@ -0,0 +1,27 @@
+//          Copyright Naoki Shibata 2010 - 2017.
```
fpetrogalli-arm: Why splitting this into main file and separate file? There is no I/O going on here as far as I can tell.

shibatch: Since tester3.c is compiled with options for enabling usage of vector extensions turned on, there is always possibility that the compiler generates instructions with that vector extension even without using intrinsics. In that case, the whole program crashes if that extension is not available on the computer. In our case, we want to make the testing succeed in that case, so the main function has to be separated and compiled without vector extension.

shibatch: It is already running on travis.

https://travis-ci.org/shibatch/sleef/jobs/317591746

src/libm-tester/CMakeLists.txt
```diff
@@ -34,8 +42,11 @@ macro(test_extension SIMD)
     add_dependencies(${TARGET_IUT${SIMD}} ${TARGET_HEADERS})
     add_dependencies(${TARGET_IUT${SIMD}} ${TARGET_LIBSLEEF})
 
-    add_test(NAME ${TARGET_IUT${SIMD}}
-      COMMAND ${TARGET_TESTER} ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${TARGET_IUT${SIMD}})
+    list(FIND SKIP_TEST ${SIMD} INDEX_SKIP_TEST)
+    if (INDEX_SKIP_TEST EQUAL -1)
```
fpetrogalli-arm: What is going on here? Why do you want to turn off the regular tester when you are building the tester3 executables? Is this something you just did for debug, to skip running the iutsimd tests because it takes too long for them to complete?

shibatch: Since tester3 checks if the computation of the given two implementations are all same, it should be enough to check only one of them using the original tester. I am going to add more implementations with different bit lengths, and I think testing will take too much time without doing like this.

shibatch: As you see, there is an option SLEEF_TEST_ALL_IUT. By turning this on, all implementations are checked by the original tester. I believe it is reasonable to test all of them only occasionally.

shibatch:
@fpetrogalli-arm Do you still have any concern in introducing tester3?
This implementation of tester3 only checks small arguments for trigonometric functions, but this limitation will be removed once the scalar functions based on SIMD implementation will be implemented.

Is it better to start discussing development plan for version 3.3?
I think the plan towards 3.3 will be pretty complicated.

Issue #101, opened 2017-11-10
Add log2 function


musm:
Should be straightforward to add

shibatch:
I have many patches that are ready to commit, but I am waiting for the right time to come before committing, since I need to make changes on many places. I will commit them after the cmake branch will be merged, and that should happen soon.

PR #130, opened 2017-12-19
Addition of log2

shibatch:
This patch adds log2 and log2f.

https://github.com/shibatch/sleef/issues/101


fpetrogalli-arm:
Could you please also add the "finite" names testing? In https://github.com/shibatch/sleef/blob/master/src/libm-tester/finite_names_testing.c

They are not complete as most of the other functions are missing, but if you add the tests for `__log2[f]_finite` here in this PR, we won't forget it for sure.

shibatch:
Since the tester only covers asin and acos at this point, I think it is better to make another PR for committing a patch to make it cover a lot more functions.
Please write an issue to explain what it is really going to test.

src/libm-tester/iutsimd.c
```diff
@@ -416,12 +416,7 @@ int do_test(int argc, char **argv) {
     func_d_d("exp10", xexp10);
     func_d_d("expm1", xexpm1);
     func_d_d("log10", xlog10);
-    func_d_d("log1p", xlog1p);
```
fpetrogalli-arm: Could you please explain why you are removing these lines?

shibatch: These lines were accidentally duplicated in the previous code.
The same tests was executed twice, which was just waste of time.

Issue #85, opened 2017-10-11
Running tests on Intel sde


shibatch:
I am considering to run the tests on Intel software development emulator. With this, we can test AVX2 and AVX512F implementations in addition to the currently tested ones.

Testing won’t be too much slower than now, since only IUT runs on the emulator.

We can see a similar travis configuration in the following pull request.

https://github.com/flame/blis/pull/124

shibatch:
I checked the license of sde, and I found that redistribution is not allowed. The package cannot be downloaded directly from Intel's server since we need to agree the license. So, we need to put the package at somewhere else, and download the package to the CI server. I am not fully sure if this does not infringe the license.

jeffhammond:
I downloaded the install file and put it in Dropbox with the location hidden by Travis secure environment variables. 

When I talked to SDE team about this, they were trying to enforce license acceptance on users. You’ll accept license when you download so I don’t think it’s a problem as long as you’re the only person running the SDE binary you download. However, I can ask if you think it’s necessary.

shibatch:
Hi Jeff,
Since everything is visible publicly, I don't want to do anything suspicious. I guess there is demand by people other than us. Hiding a download location by a secure environment variable is a nice idea, and it sounds not infringing the license. I think it would be easier for people if usage on CI environments is clearly stated in the license.

@fpetrogalli-arm How do you think? If it is okay for you, I will start implementing it.

jeffhammond:
I will see what I can do to address this.

shibatch:
Thank you so much. Please also tell them to allow using Intel Compiler on CIs. 🙂

jeffhammond:
Write me an email (GitHub shows it) and we can discuss. That’s a problem I
know how to solve.
-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

fpetrogalli-arm:
@shibatch : of course it is OK for me, and I am looking forward to see it execute on CI!
@jeffhammond : thanks for reaching out!

PR #131, opened 2017-12-22
Running tests on intel sde

shibatch:
This patch makes the testers to run on Intel SDE.
The SDE is only used when the corresponding extension is not available on the computer.

Travis terminates running programs if it does not receive anything from stdout of the program for 10 minutes. So this patch uses travis_wait to extend the timeout to 30 minutes. As a result, the output from the tester cannot be seen until it finishes.

https://github.com/shibatch/sleef/issues/85


src/libm-tester/CMakeLists.txt
```diff
@@ -37,9 +37,14 @@ macro(test_extension SIMD)
     add_dependencies(${TARGET_IUT${SIMD}} ${TARGET_HEADERS})
     add_dependencies(${TARGET_IUT${SIMD}} ${TARGET_LIBSLEEF})
 
-    add_test(NAME ${TARGET_IUT${SIMD}}
-      COMMAND ${TARGET_TESTER} ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${TARGET_IUT${SIMD}})
-    
+    if (SDE_COMMAND)
+      add_test(NAME ${TARGET_IUT${SIMD}}
+	COMMAND ${TARGET_TESTER} "--sde" ${SDE_COMMAND} ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${TARGET_IUT${SIMD}})
+    else()
```
fpetrogalli-arm: Please use the form:

```
if (SDE_COMMAND)
// set SDE_RUN to be "--sde" ${SDE_COMMAND}
else()
// set SDE_RUN to be "" (empty)
endif()
add_test(NAME ${TARGET_IUT${SIMD}}
COMMAND ${TARGET_TESTER} ${SDE_RUN} ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${TARGET_IUT${SIMD}})
```

You could move the if statement at global configuration when detecting the SDE_COMMAND.

src/libm-tester/tester.c
```diff
@@ -4573,14 +4575,19 @@ int main(int argc, char **argv) {
   for(a2s=1;a2s<argc;a2s++) {
     if (strcmp(argv[a2s], "--flushtozero") == 0) {
       enableFlushToZero = 1;
+    } else if (a2s+1 < argc && strcmp(argv[a2s], "--sde") == 0) {
```
fpetrogalli-arm: Is the "--sde" option really needed? I am not really keen in having simulator specific options in the tester executable source code.

How about adding a wrapper script around the SDE_COMMAND that allows to invoke tester and the iutsimd executable as follows?
```
/path/to/tester /path/to/run_with_simulator.sh /path/to/iutWhateverSimd
```

Run with simulator is something I already have for SVE, it accepts one parameter in input and execute it via the simulator. Here what we could do to make it "simulator independent":

```
#!/bin/bash
SIMULATOR=$1
IUT=$2
./SIMULATOR ./IUT
```

I think that by doing so we could avoid all the changes you have done in iutsimd.c and tester.c, and the logic for handling the different simulators invocation should be less verbose in the bash script.

.travis.yml
```diff
@@ -35,6 +35,9 @@ matrix:
         - MATRIX_EVAL="CC=clang-5.0 && CXX=clang++-5.0"
 
 before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then wget -q ${SDE_URL}                            ; fi
```
fpetrogalli-arm: Where does SDE_URL come from? Is it a Travis env variable?

src/libm-tester/tester.c
shibatch: The reason why I made this option is that the emulation is slow. It takes 12 minutes to test with sde, while it only takes 2 minutes without sde. With this option, the tester can check if sde is needed, and use it only if it is needed.

https://travis-ci.org/shibatch/sleef/jobs/319800736

.travis.yml
shibatch: SDE_URL is a Travis secure environment variable. Since redistribution of SDE is not allowed, we need to hide where we are downloading the SDE package from.

https://github.com/shibatch/sleef/issues/85

src/libm-tester/tester.c
fpetrogalli-arm: I see, it make sense. Thanks for the explanation.

fpetrogalli-arm:
Wait for CI to finish, then squash all commits.

PR #132, opened 2017-12-22
[cmake] Optional build of static libs

shibatch:
This patch allows optional builds of static libs.
On Windows, SLEEF_STATIC_LIBS macro has to be defined before including sleef.h and sleefdft.h in order to link the user code with the static libs.
The GNUABI version of libraries are built as shared libraries regardless of the option.
This patch also includes small modification of CMakelists.txt that allows build on Cygwin.


CMakeLists.txt
```diff
@@ -21,6 +29,7 @@ endif()
 
 # Set output directories for the library files
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
```
fpetrogalli-arm: This doesn't seem to be used anywhere, please remove.

src/libm-tester/CMakeLists.txt
```diff
@@ -24,21 +25,28 @@ set(IUT_LIST ${TARGET_IUT})
 
 set(IUT_SRC iutsimd.c iutsimdmain.c testerutil)
 
+if (CMAKE_SYSTEM_NAME MATCHES "CYGWIN")
```
fpetrogalli-arm: Why do you need to skip these tests? It seems to me that you are building them, even on cygwin, (see line 35), but then you decide not to run them. Aren't the iuts just supposed to bail out if the platform does not support the vector architecture? Or am I missing something?

Having said that, I would prefer not to mix different aspects of the build system in the same pull request. I would prefer you to move the cygwin specific bits from here to a new pull request and limit this one to SLEEF_STATIC_LIBS.

CMakeLists.txt
shibatch: This is implicitly used by CMAKE, and it is for specifying the output directory for static libs.

PR #134, opened 2017-12-22
[build] Don't force building shared libraries

mkbosmans:
By not explicitly using add_library(SHARED), the user can override the default
and choose to build libsleef as a static libraries by setting SLEEFLIBM_STATIC=ON.

There are still two points to be addressed later:
 - The dft library is always built as a shared library.
 - On Windows when building sleeflibm as a static library,
   the imported symbols in dft.dll are wrong,
   causing a linker error.


shibatch:
Please check https://github.com/shibatch/sleef/pull/132.

mkbosmans:
Ah, that's indeed a more comprehensive solution.
It looks OK, I'll test it next week.

Issue #133, opened 2017-12-22
Dispatcher chooses AVX function on AVX2 hardware


mkbosmans:
Since #121 is fixed, the dispatch functions are compiled in libsleef again.
But now I see the problem that the dispatch function chooses AVX instead of AVX2.

For a certain (fixed) number sincos evaluations I see the following runtime:
  Sleef_sincosf8_u35avx:  74 ms
  Sleef_sincosf8_u35avx2: 54 ms
  Sleef_sincosf8_u35:     80 ms

This is not what I would expect.
Indeed, when profiling the code, the AVX function is being called.

PR #135, opened 2017-12-23
[dispatcher] Fix dispatcher

shibatch:
This patch fixes the problem in which the dispatchers do not choose the best available extension.

https://github.com/shibatch/sleef/issues/133

In order to prevent a similar bug to be made again, the selected extension will be displayed upon execution of iutdsp128 and iutdsp256.

### DIFF
--- a/src/libm/dispavx.c.org
+++ b/src/libm/dispavx.c.org
@@ -138,4 +138,28 @@ static int cpuSupportsFMA4() {
   static CONST fptype (*pfn)(fptype arg0, fptype arg1, fptype arg2) = dfn; \
   EXPORT CONST fptype funcName(fptype arg0, fptype arg1, fptype arg2) { return (*pfn)(arg0, arg1, arg2); }
 
+#define DISPATCH_si_si(funcName, pfn, dfn, funcavx, funcfma4, funcavx2) \
+  static CONST int (*pfn)(int arg0);					\
+  static CONST int dfn(int arg0) {					\
+    int CONST (*p)(int) = funcavx;					\
+    SUBST_IF_FMA4(funcfma4);						\
+    SUBST_IF_AVX2(funcavx2);						\
+    pfn = p;								\
+    return (*pfn)(arg0);						\
+  }									\
+  static CONST int (*pfn)(int arg0) = dfn;				\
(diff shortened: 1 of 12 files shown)


src/libm/dispavx.c.org
```diff
@@ -138,4 +138,28 @@ static int cpuSupportsFMA4() {
   static CONST fptype (*pfn)(fptype arg0, fptype arg1, fptype arg2) = dfn; \
   EXPORT CONST fptype funcName(fptype arg0, fptype arg1, fptype arg2) { return (*pfn)(arg0, arg1, arg2); }
 
+#define DISPATCH_si_si(funcName, pfn, dfn, funcavx, funcfma4, funcavx2) \
```
fpetrogalli-arm: Can you explain what DSPATCH_si_si and DISPATCH_sp_si names are for? In particular, what si and sp stand for?

src/libm/dispavx.c.org
```diff
@@ -138,4 +138,28 @@ static int cpuSupportsFMA4() {
...
+  static CONST int (*pfn)(int arg0);					\
+  static CONST int dfn(int arg0) {					\
+    int CONST (*p)(int) = funcavx;					\
+    SUBST_IF_FMA4(funcfma4);						\
+    SUBST_IF_AVX2(funcavx2);						\
```
fpetrogalli-arm: If I understand correctly, you assume that AVX2 is always better than FMA4, and FMA4 is always better than AVX?

src/libm/dispavx.c.org
shibatch: There is not much meaning. si is int, sp is void *.

src/libm/dispavx.c.org
shibatch: Yes.

src/libm/dispavx.c.org
fpetrogalli-arm: The names don't make it clear. I think it is worth adding a comment at the top of the sources saying what the DISPATCH_A_B macro mean, for each of the values of A and B.

shibatch: Actually I wanted to rename _d to _vd or _vf, and use _i for int. I will also do this.

PR #138, opened 2017-12-29
[cmake] Reliable parallel build

shibatch:
A parallel build with cmake has a known problem in which ADD_CUSTOM_COMMAND is executed multiple times. Because of this, parallel build is not reliable especially on Visual Studio.

https://cmake.org/pipermail/cmake/2008-October/024492.html

In this patch, the generated code is once written to a file with the PID of the process after it, and then that file is renamed to the specified file name after the generation is done. In this way, the other processes do not see incomplete contents of the generated file.


fpetrogalli-arm:
I think you should create the correct dependencies as suggested in https://cmake.org/pipermail/cmake/2008-October/024492.html instead of using the `copycat` executable. You could use temporary header files for each SIMD target, build up a list of them and concatenate them to the final header file.

shibatch:
I tried the way suggested in the page, but I can't make it work.
It works on Linux, but on MSVC, add_custom_command is still executed many times.
It is not always possible to make add_custom_target drive the command and make the other targets depend on that one, since other targets depends on the generated source files.
Please tell me if there is a clean way without the copycat tool.
At this point, copycat tool seems an easy way to make parallel build reliable.

fpetrogalli-arm:
What did you exactly try? Can you post a patch of what you were doing to resolve the dependencies?

Let see if we are on the same page here.

Say that sleef.h depends of the headers generated for X and Y SIMD variants.

You could create targets for temporary files tmpX.h and tmpY.h, and make them a requirement for sleef.h:

```

add_custom_command(tmpX.h COMMAND mkrename ...)
add_custom_command(tmpY.h COMMAND mkrename ...)
set (LIST_HOLDING_TMP_FILES tmpX.h tmpY.h)
add_custom_command(sleef.h COMMAND cat ${LIST_HOLDING_TMP_FILES} >> sleef.h
  DEPENDS  ${LIST_HOLDING_TMP_FILES})
```

Is this similar to what you have tried? Of course, it should be made a loop to loop on all possible vector extensions.

shibatch:
I made a branch. Please see what happens when it is built in parallel on Visual Studio.
It's not sleef.h, but rename*.h.

https://github.com/shibatch/sleef/tree/Reliable_parallel_build_without_copycat

shibatch:
The same behavior can be reproduced by the following CMakelists.txt.
Again, this works on Linux, but those headers are generated multiple times on Windows.

```
foreach(N 0 1 2 3 4 5 6 7 8 9)
  set(GEN${N} ${CMAKE_CURRENT_BINARY_DIR}/generated${N}.c)
  add_custom_command(OUTPUT generated${N}.c COMMAND echo void func${N}'(){}' > ${GEN${N}})
  set(GENERATED ${GENERATED} ${GEN${N}})
endforeach()

add_custom_target(files_generated DEPENDS ${GENERATED})

foreach(N 0 1 2 3 4 5 6 7 8 9)
  add_library(lib${N} STATIC ${GENERATED})
  add_dependencies(lib${N} files_generated)
endforeach()
```

On Windows, please replace '(){}' with (){}

fpetrogalli-arm:
@xoofx , do you have an idea of what is the problem here?

xoofx:
@shibatch  If you add ALL to the add_custom_target of your previous example, it should work:

```
add_custom_target(files_generated ALL DEPENDS ${GENERATED})
```

Haven't dig into why sleef is not reliable with parallel build, but if the headers must be generated before proceeding, the dependent targets should be dependent on `${TARGET_HEADERS}`

shibatch:
@xoofx I tried it, but it didn't work.

https://www.dropbox.com/s/vqd97q4ffks2a1x/msvclog.txt?dl=0

xoofx:
Hm, it seems that removing the `generated*.c` files manually and re-running with maxcputcount works, but not if you start from a clean folder (where the dependent CMakeFiles are not yet generated)... so it looks like a msbuild bug somehow...

Don't have a simple workaround for this, so if the copycat is working, we can keep it. Maybe add a comment in the CMakeLists.txt to explain why it is done like this (to workaround a bug with msvc parallel build)

shibatch:
I found another thread. This seems to be a known issue.

https://cmake.org/pipermail/cmake/2014-February/057025.html

fpetrogalli-arm:
Do we really need parallel builds on windows? To my understanding, the patch with the correct dependencies in the CMakeLists.txt in https://github.com/shibatch/sleef/tree/Reliable_parallel_build_without_copycat works on Linux/OSX. My preference would be to avoid adding the copycat code and just add a comment saying that the parallel build are not reliable on windows because of a build system problem, and build serially on appveyor. To me this is just a build problem, so no big deal, as we rely more on parallelism for testing, not for building.

shibatch:
Okay, I will remove copycat.

src/libm/CMakeLists.txt
```diff
@@ -194,7 +195,7 @@ if (SLEEF_ARCH_X86)
     COMMENT "Generating dispsse.c"
     COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/dispsse.c.org ${CMAKE_CURRENT_BINARY_DIR}/dispsse.c
     COMMAND ${TARGET_MKDISP} 2 4 __m128d __m128 __m128i sse2 sse4 avx2128 >> ${CMAKE_CURRENT_BINARY_DIR}/dispsse.c
-    DEPENDS ${TARGET_MKDISP}
+    DEPENDS ${TARGET_MKDISP} ${TARGET_COPYCAT}
```
fpetrogalli-arm: I believe this `${TARGET_COPYCAT}` is a leftover from the previous version. Please remove.

PR #139, opened 2018-01-04
[GNUABI] Adding masked versions of functions

shibatch:
This patch will add masked versions of the functions to the GNU ABI version of the library.


src/libm/sleefsimddp.c
```diff
@@ -2783,4 +2783,7 @@ EXPORT CONST vdouble __sinh_finite     (vdouble)          __attribute__((weak, a
 EXPORT CONST vdouble __sqrt_u05_finite (vdouble)          __attribute__((weak, alias(str_xsqrt_u05 )));
 EXPORT CONST vdouble __tgamma_u1_finite(vdouble)          __attribute__((weak, alias(str_xtgamma_u1)));
 
+#ifdef HEADER_MASKED
+#include HEADER_MASKED
```
fpetrogalli-arm: @shibatch , thank you for working on this. 
I'd prefer having the explicit functions here, with a "masked" name being generated by the `mkrename_gnuabi` tool.

For example:
```
#ifdef ENABLE_MASKED_INTERFACE
vdouble xcos_masked(vdouble, vopmask) {
  // function body
  return xcos(vdouble); // temporary implementation
}
// ...
#endif

```
The reason for doing so is that for now we are just using the original unmasked functions to execute the masked version, but in the future we will come up with fully-masked implementations, using masked intrinsics. It will be easier to extend this work than to deal with the header file.

Also, I think we should enable the masked symbols only for those vector extension that support masking natively, i.e. AVX512F.

shibatch: If we implement in that way, we need to maintain many hand-written functions, which would be troublesome. I think automatically generating the header is easier to maintain and less bug-prone.

Also, it is unclear when we will start implementation of the body of masked functions. It is not too late to convert the header to fully-implemented functions at that time.

fpetrogalli-arm: I disagree with your "easier to maintain and less bug-prone".

The thing is that the masked functions will probably have to be replaced with a proper implementation as soon as someone start to use them, because they will be not as good as they could be when implemented with native predication. I also expect such replacement to happen gradually, function by function, not with a single commit: in that case, it will be easier to modify the functions in place instead of having to deal with the auto generative process.

Configure.cmake
```diff
@@ -98,6 +98,17 @@ command_arguments(RENAME_PARAMS_GNUABI_AVX2    avx2 d 4 8 __m256d __m256 __m128i
 command_arguments(RENAME_PARAMS_GNUABI_AVX512F avx512f e 8 16 __m512d __m512 __m256i __m512i __AVX512F__)
 command_arguments(RENAME_PARAMS_GNUABI_ADVSIMD advsimd n 2 4 float64x2_t float32x4_t int32x2_t int32x4_t __ARM_NEON)
 
+command_arguments(MKMASKED_PARAMS_GNUABI_SSE2_dp    sse2 b 2)
+command_arguments(MKMASKED_PARAMS_GNUABI_SSE2_sp    sse2 b -4)
+command_arguments(MKMASKED_PARAMS_GNUABI_AVX_dp     avx c 4)
+command_arguments(MKMASKED_PARAMS_GNUABI_AVX_sp     avx c -8)
+command_arguments(MKMASKED_PARAMS_GNUABI_AVX2_dp    avx2 d 4)
+command_arguments(MKMASKED_PARAMS_GNUABI_AVX2_sp    avx2 d -8)
+command_arguments(MKMASKED_PARAMS_GNUABI_AVX512F_dp avx512f e 8)
```
fpetrogalli-arm: I think we should enable the masked interface only for AVX512F for now. The extensions with no predication are unlikely to benefit from it.

src/libm/mkmasked_gnuabi.c
```diff
@@ -0,0 +1,120 @@
...
+    case 0:
+      printf("EXPORT CONST %s %s(%s a0, vopmask m) { return %s(a0); }\n",
+	     vfpname[fptype], funcname[1], vfpname[fptype], funcname[0]);
+      break;
+    case 1:
+      printf("EXPORT CONST %s %s(%s a0, %s a1, vopmask m) { return %s(a0, a1); }\n",
+	     vfpname[fptype], funcname[1], vfpname[fptype], vfpname[fptype], funcname[0]);
+      break;
+    case 2:
+      if (fptype == 0) {
```
fpetrogalli-arm: I would change this to have a `sizeoffp[fptype] == sizeof(double)` as the test, to make it explicit for the reader.

src/libm/mkmasked_gnuabi.c
```diff
@@ -0,0 +1,120 @@
...
+	printf("EXPORT void %s(vdouble a0, double *a1, double *a2, vopmask m) {\n", funcname[1]);
+	printf("  double s[VECTLENDP], c[VECTLENDP];\n");
+	printf("  int32_t mbuf[VECTLENSP];\n");
+	printf("  %s(a0, s, c);\n", funcname[0]);
+	printf("  vstoreu_v_p_vi2(mbuf, vcast_vi2_vm(vand_vm_vo64_vm(m, vcast_vm_i_i(-1, -1))));\n");
+	printf("  for(int i=0;i<VECTLENDP;i++) {\n");
+	printf("    if (mbuf[i*2]) { *a1++ = s[i]; *a2++ = c[i]; }\n");
+	printf("  }\n");
+	printf("}\n");
+      } else {
```
fpetrogalli-arm: And I wouldn't use and else here. another if statement, checking that the size is `sizeof(float)`.

And after such if block I would add an assertion like
```
assert(0 && "Invalid size of FP data");
```
to make sure we don't end up introducing bogus stuff.

src/libm/sleefsimddp.c
fpetrogalli-arm: Update: I am happy with the idea of the auto-generated interface. We will work on a new version according to #142 .

fpetrogalli-arm:
I am creating a merge commit instead of squashing because there are commits from master branch which for merged in this PR.

PR #140, opened 2018-01-05
[aarch64] Fix fmin and fmax on aarch64

shibatch:
vmaxnum_v*_v*_v* and vminnum_v*_v*_v* are helper functions for utilizing instructions for choosing larger and smaller elements from two given vectors, and they are introduced in the following PR.

https://github.com/shibatch/sleef/pull/109

It is said that vmaxnmq and vminnmq on aarch64 are IEEE754-conformant, but I recently found that handling of signalling NaN by these instructions is not conforming to the specification of fmin and fmax functions in the ANSI C standard.

This patch fixes that problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then wget -q ${SDE_URL}                            ; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then tar xf sde-external-8.12.0-2017-10-23-lin.tar.bz2; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then export PATH=${PATH}:${TRAVIS_BUILD_DIR}/sde-external-8.12.0-2017-10-23-lin; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo rm /etc/apt/sources.list.d/mongodb-3.2.list; fi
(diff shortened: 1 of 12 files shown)


fpetrogalli-arm:
Would it be possible to add a regression test? Also, do you mind linking some official information about this problem?
Moreover, we introduced the `SLEEF_SINGLE_MINMAXNUM_AVAILABLE`and `SLEEF_DOUBLE_MINMAXNUM_AVAILABLE` in #109. If we opt for the solution in this PR, I think we should remove those macros and all the changes related to that.

shibatch:
Do you think it is better to test a signalling NaN input to all functions?

fpetrogalli-arm:
> Do you think it is better to test a signalling NaN input to all functions?

No, I think it is enough to add a regression test for this particular PR.

shibatch:
apt-get update at travis did not work anymore, so I added lines for updating expired keys and removing an apt source relating to mongo-db following the comments in https://github.com/travis-ci/travis-ci/issues/8554 to .travis.yml.

shibatch:
apt-get update problem seems to be fixed now.

https://github.com/travis-ci/travis-ci/issues/9037

shibatch:
As for the operation carried out by the aarch64 instructions, they are the minNum and maxNum operations defined in section 5.3 and 6.2.3 of IEEE Std 754™-2008. 

http://ieeexplore.ieee.org/document/4610935/

The IEEE standard specifies that signaling NaNs should be propagated in these operations.

On the other hand, the C99 standard does not distinguish between a quiet NaN and a signaling NaN.

Since SLEEF is a substitute of the C99 math library, propagation of a signaling NaN is not desirable.

PR #141, opened 2018-01-05
[cmake] Build SLEEF GNUABI if aliases are available.

fpetrogalli-arm:


fpetrogalli-arm:
`libsleefgnuabi.so` installed on linux: https://travis-ci.org/shibatch/sleef/jobs/325396050#L5567

PR #146, opened 2018-01-12
[Makefile] Remove makefiles

shibatch:
This patch removes reduces functionality of makefiles. They are not removed since we still need them for cleaning up.

The location of sleefdft.h is changed from include to src/dft/include. CMakefiles are changed accordingly.


shibatch:
The reason that I changed CMakeLists.txt is that the location of sleefdft.h is changed from include to src/dft/include.

I want to keep makefiles for cleaning up files like *~ and *.o.

shibatch:
*~ is generated when I edit files.
Usually I have a.out and *.s files in the directories.
Makefiles are not only for building, but there are other uses.

fpetrogalli-arm:
> *~ is generated when I edit files.
> Usually I have a.out and *.s files in the directories.

yeah, I still think you should add a `.gitignore` file listing all what has to be considered junk in the source folder, not craft makefiles around that.

> Makefiles are not only for building, but there are other uses.

Not under  the `src` folder. Am I missing something?

shibatch:
I don't know how you edit and debug the programs, but in my style, I make lots of junk files under src directory.
Don't you want to remove junk files under src directory?

fpetrogalli-arm:
> I don't know how you edit and debug the programs, but in my style, I make lots of junk files under src directory.

Of course I do have a lot of junk in the tree of my project. we are all in the same boat! :)

> Don't you want to remove junk files under src directory?

Of course I do: `git clean [-f]` (careful with the `-f`).

shibatch:
Rebased. The new PR is https://github.com/shibatch/sleef/pull/175.

PR #149, opened 2018-01-19
[document] Update html documents

shibatch:
This patch updates the document under html directory.


doc/html/compile.xhtml
```diff
@@ -15,67 +17,190 @@
...
+<p class="noindent">
+6. Install the library under ../my-sleef/install by running:
+</p>
+<pre>$ make install</pre>
 
-<h3>Compiling library with Microsoft Visual C++</h3>
+<p class="noindent">
+7. You can execute the tests by running:
+</p>
+<pre>$ make test-libm</pre>
```
fpetrogalli-arm: `test-libm` does not exist as a target in the cmake configuration. The target to run the tests is `make test`, but it executes and complete correctly only if all the targets are build with `make all` or just `make`.

Is this a leftover of the old docs related to the Makefile builds?

fpetrogalli-arm:
This LGTM, no need to run CI on it.

Issue #137, opened 2017-12-23
Adding aliases without vector extension name to aarch64 and aarch32


shibatch:
For the x86 vector functions, function names without vector extension is used to call the dispatcher.
In order to improve portability of the source code of the programs that call SLEEF functions, it would be also desirable for aarch64 and aarch32 functions to have aliases without vector extension name.

For example, Sleef_sind2_u10 can be used instead of Sleef_sind2_u10advsimd.

If a weak alias is not supported by the compiler that compiles SLEEF library, functions with the corresponding names are added instead.

This change will not affect the current GNUABI version of functions.

PR #153, opened 2018-01-23
[Alias] Add aliases without vector extension name (2)

shibatch:
This patch adds aliases without vector extension name to the existing functions.
This is implementation of issue https://github.com/shibatch/sleef/issues/137.
In addition to aarch64 and aarch32, this patch adds aliases to AVX512F functions.


src/libm/CMakeLists.txt
```diff
@@ -189,23 +197,90 @@ endif()
...
     # Create a library
-    add_library(${OBJECT_${SIMD}} OBJECT ${SIMD_SOURCES} ${HEADER_${SIMD}})    
-    target_compile_definitions(${OBJECT_${SIMD}} PRIVATE
-      ENABLE_${SIMD}=1
-      DORENAME=1
-      ${COMMON_TARGET_DEFINITIONS}
-      )
-    add_dependencies(${OBJECT_${SIMD}} rename${SIMD}.h.generated)
+    add_library(${OBJECTsp_${SIMD}} OBJECT sleefsimdsp.c ${HEADER_${SIMD}})    
+    if (INDEX_ALIAS EQUAL -1)
```
fpetrogalli-arm: I think we can simplify this. We should always generate the alias header file, just use the `ENABLE_ALIAS` macro in the `mkalias` tool sources. If it set to one, generated the alias, otherwise leave the file empty. This would make possible to remove this extra logic that you have added here.

shibatch: I don't fully understand what you mean, but the alias is only generated if a dispatcher is not available. For example, Sleef_sind4_u10 calls a dispatcher, while Sleef_sind8_u10 is an alias of Sleef_sind8_u10avx512f.

Currently, ENABLE_ALIAS macro is used to indicate if the compiler supports weak alias.

fpetrogalli-arm: Apologies, I should have make it more clear. I think you can avoid adding such logic in the cmake configuration.

I don't think it is necessary to split the configuration for single precision and double precision. You have done this because you have decided to simplify the mkalias tool to accept only SP or only DP parameters, but his is not what we have done for other `mk*` tools. Tools like `mkrename` accept at once all the SP and DP parameters, it is then the `ENABLE_DP` and `ENABLE_SP` macros in the code of the library that decide what to load.

src/libm/CMakeLists.txt
```diff
@@ -189,23 +197,90 @@ endif()
...
+      target_compile_definitions(${OBJECTsp_${SIMD}} PRIVATE
+	ENABLE_${SIMD}=1
+	DORENAME=1
+	${COMMON_TARGET_DEFINITIONS}
+	)
+    else()
+      add_custom_command(
+	OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/include/alias${SIMD}sp.h ${CMAKE_CURRENT_BINARY_DIR}/include/alias_${SIMDLC}_sp.h
+       	COMMENT "Generating alias_${SIMDLC}sp.h"
+       	COMMAND ${TARGET_MKALIAS} ${ALIAS_PARAMS_${SIMD}_SP} > ${CMAKE_CURRENT_BINARY_DIR}/include/alias_${SIMDLC}_sp.h
```
fpetrogalli-arm: Our internal CI are simulating the SVE version of the library, both with native aarch64 simulation of SVE and with X86 based tools. Because of the latter configuration,  I started cross compiling our downstream version on X86, targeting AArch64 Advanced SIMD and SVE. In such cases, it turned out that the `COMMAND` invocation of `add_custom_command` are not able to find the executable that are just referred via the target name, like in `COMMAND ${TARGET_MKALIAS} ...`. It turned out that the command should be specified using generator expression, as explained in https://stackoverflow.com/questions/4346281/path-to-target-output-file, resulting in  `COMMAND $<TARGET_FILE:${TARGET_MKALIAS} ...>`

shibatch:
The new branch is https://github.com/shibatch/sleef/tree/Add_aliases_without_vector_extension_name3

PR #157, opened 2018-01-26
[cmake] Cross compiling tweaks.

fpetrogalli-arm:
Some of our downstream CI pipelines rely on cross-compiling SLEEF on
X86 and targeting AArch64.

This patch constists of the changes I needed to implement in the cmak
configuration to make it work on those CI environment.

No functional change is intended.


fpetrogalli-arm:
@xoofx , do you know why `libarraymap` needs to be static for the Windows build?

shibatch:
Adding the following change to common.c will fix the failing build on OSX.
```
 #elif defined(__APPLE__)
 #include <sys/time.h>
 
-EXPORT void *Sleef_malloc(size_t z) { return malloc(z); }
+EXPORT void *Sleef_malloc(size_t z) { void *ptr = NULL; posix_memalign(&ptr, 256, z); return ptr; }
```

shibatch:
Is it possible to implement "importing the executable targets"?

https://cmake.org/Wiki/CMake_Cross_Compiling#Using_executables_in_the_build_created_during_the_build

shibatch:
Cross compilation without qemu is not hard at all. Define the following function somewhere

```
function(add_host_executable TARGETNAME SOURCE)
  if (NOT CMAKE_CROSSCOMPILING)
    add_executable(${TARGETNAME} ${SOURCE})
  else()
    add_executable(${TARGETNAME} IMPORTED)
    set_property(TARGET ${TARGETNAME} PROPERTY IMPORTED_LOCATION ${NATIVE_BUILD_DIR}/bin/${TARGETNAME})
  endif()
endfunction()
```
Then, replace add_executable that is defining a command executed on the host computer to add_host_executable.

Before cross compilation, native build has to be executed. The second cmake can be executed by the following command.
```
cmake -DCMAKE_TOOLCHAIN_FILE=../toolchain-aarch64.cmake -DNATIVE_BUILD_DIR=`pwd`/../build-x86_64 ..
```

toolchain-aarch64.cmake is like following.
```
SET (CMAKE_CROSSCOMPILING   TRUE)
SET (CMAKE_SYSTEM_NAME      "Linux")
SET (CMAKE_SYSTEM_PROCESSOR "aarch64")

SET(CMAKE_FIND_ROOT_PATH  /usr/aarch64-linux-gnu /usr/include/aarch64-linux-gnu /usr/lib/aarch64-linux-gnu)

find_program(CMAKE_C_COMPILER aarch64-linux-gnu-gcc aarch64-linux-gnu-gcc-8 aarch64-linux-gnu-gcc-7 aarch64-linux-gnu-gcc-6 aarch64-linux-gnu-gcc-5)

SET(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
SET(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY BOTH)
SET(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
```

fpetrogalli-arm:
> Adding the following change to common.c will fix the failing build on OSX
@shibatch , which failure? Travis seems happy to me, even on OSX.

> Cross compilation without qemu is not hard at all

@shibatch Thanks for the suggestion. I will add this in a separate patch.

Can I merge this one now?

shibatch:
>@shibatch , which failure? Travis seems happy to me, even on OSX.

It failed once. I restarted the job. If the job is executed on a computer with AVX, it fails.

>Can I merge this one now?

Yes

fpetrogalli-arm:
@shibatch Are these the failures you were talking about when suggesting the OSX fix?

shibatch:
yes

fpetrogalli-arm:
@shibatch , are you happy for me to submit a patch with the code you suggested to fix the OSX build?

shibatch:
yes

Issue #145, opened 2018-01-09
[GNUABI] Finite symbols tests


fpetrogalli-arm:
The test in https://github.com/shibatch/sleef/blob/master/src/libm-tester/finite_names_testing.c make sure that the finite symbols expected for GLIBC compatibility are available in the library.

They are currently checking only a couple of functions for the Intel vector extensions.

We should decide whether we want to keep them and extend them to all the finite symbols and all the supported vector extensions, or if we want to remove them.

Any thoughts?

One argument for keeping them and extending them is the following. I came up with this idea when I first used the library in our compiler. The compiler was generating the finite names (because of the `-ffast-math` option) which were not available in the library. I added them in the library and decided that I should have made a regression tests to prevent the symbols to disappear.

shibatch:
Is just checking if the symbols are exported? If so, testing using nm command would be easier to understand.

In my opinion, checking everything is desirable, but we need to come up with a good way to reduce the time taken by each testing.

If it is better to check if each function works correctly, I think modifying tester3 and checking if the functions return the same values as the non-finite functions is reasonable.

fpetrogalli-arm:
> Is just checking if the symbols are exported? If so, testing using nm command would be easier to understand.

Yes, it is just checking that the finite symbols are available (for a subset of functions). Originally the tests where written for the makefile system, and they were using the `nm` tool to check whether those symbols were available. When we ported everything to cmake, we decided with @xoofx  that C linking tests would have been better.

> In my opinion, checking everything is desirable, but we need to come up with a good way to reduce the time taken by each testing.

Linking these tests and running them takes a tiny amount of time compared to the `tester` ones.

> If it is better to check if each function works correctly, I think modifying tester3 and checking if the functions return the same values as the non-finite functions is reasonable.

It is not written anywhere that the finite calls shoudl return the same values as the non finite ones. The finite symbols are generated by a compiler  when `-ffast-math` is passed at compile time. We might need to make them faster to compete with `libmvec`, which means they will not behave the same as the non finite ones.

In general, I am keen to keep these tests, and I am also keen in expanding them not only to cover the finite GNUABI names, but the GNU ABI names in general, so that we can claim "on paper" that SLEEF is compatible with the GNUABI. 

The GNUABI names are auto-generated, I don't feel comfortable in leaving the process untested.

fpetrogalli-arm:
Here is how I intend to cover testing of the GNUABI symbols in libsleefgnuabi.

[...]

PR #162, opened 2018-01-31
New gnuabi tests

fpetrogalli-arm:
As discussed in https://github.com/shibatch/sleef/issues/145


shibatch:
I think it's better to unify how we call a simulator or emulator.

fpetrogalli-arm:
> I think it's better to unify how we call a simulator or emulator.

Of course. How about `add_sleef_test`? Or do you prefer `add_test_with_emulator`?

Also, it is still failing. I think it is because I need to make qemu aware of the library we compile. I'll sort this out.

PR #168, opened 2018-02-06
[Benchmark] New benchmarking tool with gnuplot(2)

shibatch:
This is a rebased branch of https://github.com/shibatch/sleef/pull/98.

This patch replaces the old benchmarking tool with a new tool with gnuplot.

With this new tool, the graph showing execution time can be automatically drawn.
It is now easy to see the difference between different library versions or compilers.
This tool is used to draw the graphs on sleef.org web site.

This patch implements the tool described in https://github.com/shibatch/sleef/issues/9.

A makefile is used In this patch, and there are a few reasons.
* It needs to compile the tools for SVML. These tools are independent of the other part, and thus it does not make sense to build these tools using the cmake during a build of library.
* This makefile is used as a script for measurement. It is handy since ProcData.java has to be also compiled.


src/Makefile
```diff
@@ -5,6 +5,7 @@ clean :
 	+"$(MAKE)" --directory=./gencoef clean
 	+"$(MAKE)" --directory=./libm clean
 	+"$(MAKE)" --directory=./libm-tester clean
+	+"$(MAKE)" --directory=./libm-benchmarks clean
```
fpetrogalli-arm: Remove, we don't care about makefiles.

src/libm-benchmarks/Makefile
```diff
@@ -0,0 +1,6 @@
+.PHONY: clean
```
fpetrogalli-arm: Remove, we don't care about makefiles.

src/libm-benchmarks/plot-scripts/Makefile
```diff
@@ -0,0 +1,110 @@
+BENCHSLEEF := $(shell command -v benchsleef 2> /dev/null)
```
fpetrogalli-arm: Remove, no makefiles.

src/libm-benchmarks/plot-scripts/benchsvml.c
```diff
@@ -0,0 +1,153 @@
+//          Copyright Naoki Shibata 2010 - 2017.
```
fpetrogalli-arm: This looks like a benchmark, why is it in the plot folder? The plot-scripts folder should just contain the scripts that are used to make the plots.

src/libm-benchmarks/plot-scripts/benchsvml128.c
```diff
@@ -0,0 +1,137 @@
+//          Copyright Naoki Shibata 2010 - 2017.
```
fpetrogalli-arm: Another benchmark that should sit one folder up?

src/libm-benchmarks/plot-scripts/benchsvml256.c
```diff
@@ -0,0 +1,140 @@
+//          Copyright Naoki Shibata 2010 - 2017.
```
fpetrogalli-arm: Same, benchmark code, not script for plotting.

src/libm-benchmarks/plot-scripts/benchsvml512.c
```diff
@@ -0,0 +1,137 @@
+//          Copyright Naoki Shibata 2010 - 2017.
```
fpetrogalli-arm: Same, benchmark, not script.

src/libm-benchmarks/plot-scripts/measure.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+echo
+read -p "Enter label of measurement(e.g. My desktop PC) : " label
+
+if [ -f counter.txt ]
+then
+    counter=`cat counter.txt`
```
fpetrogalli-arm: Does counter need any particular syntax?

src/libm-benchmarks/plot-scripts/measure.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
fpetrogalli-arm: It is not clear how to use this script. Who is invoking the java tool?

src/libm-benchmarks/plot-scripts/Makefile
shibatch: Did you even read the comment at the top of this page?

fpetrogalli-arm: Yes, I still think we shouldn't have makefiles down inside the tree of `src`.

If you really want to keep the makefiles, then I think you should move all this benchmarks and relative scripts at the same level of `src`, in a folder like `svml-benchmarking`, and keep it as a sort of independent project inside SLEEF.

shibatch: Originally, the whole directory including all the C codes were like an independent project.
I think that is easier to understand.
Why do you think C codes have to be compiled by cmake?

fpetrogalli-arm:
> Originally, the whole directory including all the C codes were like an independent project.
> I think that is easier to understand.
Agree. Please move it at the same level of `src`.

> Why do you think C codes have to be compiled by cmake?
I don't think it anymore. Please remove any cmake configuration related to these benchmark.

PR #170, opened 2018-02-15
[cmake] Link librt explicitly.

fpetrogalli-arm:
This is required to prevent failures in our downstream builders
when performing the 2-step cross compilation process.

Patch from Sander de Smalen @ Arm.


src/libm-tester/CMakeLists.txt
```diff
@@ -10,6 +10,11 @@ if(NOT LIB_MPFR)
   find_program(TESTER_COMMAND tester)
 endif(NOT LIB_MPFR)
 
+find_library(LIBRT rt)
+if (NOT LIBRT)
+  set(LIBRT "")
+endif()
```
shibatch: Could you move this to Configure.cmake?

PR #176, opened 2018-02-22
[cmake] Add preliminary support for MinGW

shibatch:
This patch adds preliminary support for MinGW.
Only building static libm is possible at this time.


fpetrogalli-arm:
Would it be possible to enable this on Appveyor? Or is it already enabled?

shibatch:
I added build-only testing on Appveyor.

PR #189, opened 2018-05-03
Make building tests optional

cpuhrsch:
This patch adds the following three CMAKE build option.

* BUILD_TESTS : Turned on by default. libm-tester and dft-tester will be built.


shibatch:
Why do you need this patch? If my understanding is correct, tests of the external project are defined only if such options are specified to ExternalProject_Add.

shibatch:
Okay, please sort out other problems. I will merge this patch after the PR for pytorch is approved.

shibatch:
I think this PR can be closed now.

cpuhrsch:
Yes, thank!

Issue #26, opened 2017-05-19
add exp10 function


musm:
pretty trivial to add since all the pieces are already there

shibatch:
exp10 is already there.
I think what is lacking is log2?

musm:
ah missed that, yes

musm:
btw exp2 is pretty slow compared to openlibm which just uses a table since  the exponentiation is very small

shibatch:
I want to see statistics on how frequently each function is used.

Issue #24, opened 2017-05-04
exp(1.0) not correctly rounded


musm:
For double precision sleef gives
2.7182818284590455
instead of 
2.718281828459045
despite the fact that the sleef answer is within 1 ulp, since it is such an important number it would be good that exp(1.0) was correctly rounded

shibatch:
I thought that this could be easily fixed by tweaking the coefficients, but no success at this time. I will continue trying, but this could take time.

musm:
I tried too but couldn't get any better...

BTW I have benchmarked sleef's exp function to openlibm/musm version
and the msum version is faster than sleef by usually at  around 6-11%

the sleef version is about 2x slower for small arguments since there is no explicit branch here
other cases it's usually about 0.92  slower 


See:
```julia
julia> @benchmark Base.exp(0.5)
BenchmarkTools.Trial:
  --------------
  minimum time:     9.475 ns (0.00% GC)
  median time:      10.264 ns (0.00% GC)
  mean time:        11.109 ns (0.00% GC)
  maximum time:     148.034 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark Sleef.exp(0.5)
BenchmarkTools.Trial:
  --------------
  minimum time:     12.632 ns (0.00% GC)
  median time:      14.211 ns (0.00% GC)
  mean time:        14.394 ns (0.00% GC)
  maximum time:     52.108 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark Base.exp(100.0)
BenchmarkTools.Trial:
  --------------
  minimum time:     11.842 ns (0.00% GC)
  median time:      12.238 ns (0.00% GC)
  mean time:        13.429 ns (0.00% GC)
  maximum time:     187.510 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark Sleef.exp(100.0)
BenchmarkTools.Trial:
  --------------
  minimum time:     12.632 ns (0.00% GC)
  median time:      13.027 ns (0.00% GC)
  mean time:        14.074 ns (0.00% GC)
  maximum time:     67.504 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000


julia> @benchmark Base.exp(10.0)
BenchmarkTools.Trial:
  --------------
  minimum time:     11.842 ns (0.00% GC)
  median time:      12.238 ns (0.00% GC)
[...]

PR #195, opened 2018-05-27
Add VSX support

shibatch:
This patch adds support for POWER VSX.
clang-5.0 and later is supported at this time.
gcc-8 should be able to compile the VSX code, but I haven't checked.

@avmgithub Please check if this works on your environment.


avmgithub:
@shibatch ok , will do,  do you which unit test in pytorch exercises sleef ?

shibatch:
@avmgithub I don't fully understand you mean, but please check if ctest in sleef passes.
I only checked with binfmt_misc + qemu, so I'm not confident if this works on other environments.
For PyTorch, I haven't done any test. Do you know what the standard set of tests is in a case like this?

avmgithub:
@shibatch I cloned and checkedout commit ca35e1a460df8cf3f2e9f23bef51980b5d187625   .  I hope that is the correct one.   Here were my steps:
cd sleef
git checkout ca35e1a460df8cf3f2e9f23bef51980b5d187625
mkdir build
cd build 
cmake ..
make  VERBOSE=1   <<<<   this failed stating I need to add -std=c99 .   I had to add them into 2 places .
first in line 193 ;  set(FLAGS_FASTMATH "-std=c99 -ffast-math")      and the other in 236  ; set(SLEEF_C_FLAGS "-std=c99 " "${FLAGS_WALL} ${FLAGS_STRICTMATH} ${FLAGS_NO_ERRNO}")    of Configure.cmake.

here is the error:
/home/freddie/projects/sleef/src/libm/mkmasked_gnuabi.c:58:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
   for(int i=0;funcList[i].name != NULL;i++) {
   ^
/home/freddie/projects/sleef/src/libm/mkmasked_gnuabi.c:58:3: note: use option -std=c99 or -std=gnu99 to compile your code


Once I corrected the error above.  All built fine.  I then ran "make test"
It gave me errors in the last 8 tests  (see below)
......
18/26 Test #18: roundtriptest2ddp_4_4 ............***Failed   10.44 sec
      Start 19: roundtriptest2ddp_8_8
19/26 Test #19: roundtriptest2ddp_8_8 ............***Failed    0.53 sec
      Start 20: roundtriptest2ddp_10_10
20/26 Test #20: roundtriptest2ddp_10_10 ..........***Failed    0.21 sec
      Start 21: roundtriptest2ddp_5_15
21/26 Test #21: roundtriptest2ddp_5_15 ...........***Failed    0.30 sec
      Start 22: roundtriptest2dsp_2_2
22/26 Test #22: roundtriptest2dsp_2_2 ............   Passed    0.16 sec
      Start 23: roundtriptest2dsp_4_4
23/26 Test #23: roundtriptest2dsp_4_4 ............***Failed    8.49 sec
      Start 24: roundtriptest2dsp_8_8
24/26 Test #24: roundtriptest2dsp_8_8 ............***Failed    0.50 sec
      Start 25: roundtriptest2dsp_10_10
25/26 Test #25: roundtriptest2dsp_10_10 ..........***Failed    0.20 sec
      Start 26: roundtriptest2dsp_5_15
26/26 Test #26: roundtriptest2dsp_5_15 ...........***Failed    0.30 sec

69% tests passed, 8 tests failed out of 26

Total Test time (real) =  29.56 sec

The following tests FAILED:
         18 - roundtriptest2ddp_4_4 (Failed)
         19 - roundtriptest2ddp_8_8 (Failed)
         20 - roundtriptest2ddp_10_10 (Failed)
         21 - roundtriptest2ddp_5_15 (Failed)
         23 - roundtriptest2dsp_4_4 (Failed)
         24 - roundtriptest2dsp_8_8 (Failed)
         25 - roundtriptest2dsp_10_10 (Failed)
         26 - roundtriptest2dsp_5_15 (Failed)
Errors while running CTest
make: *** [test] Error 8

shibatch:
Your compiler is too old. Try using clang-5 or later, or gcc-8.
VSX intrinsics are fully supported only by recent compilers.

avmgithub:
I'll see what I can do, unfortunately most of ppc64le users use the OS distro gcc.   Right now we're supporting RH 7.4 which comes with gcc 4.8.      Even if I switch to the latest  Advanced Toolkit for ppc64le it only goes to gcc 7.3.1  and that becomes a problem with pytorch as pytorch only seem to support gcc 6.

shibatch:
I see, but I think it's basically impossible to support VSX with gcc 4.8.
We need to wait until newer compilers are supported.

src/libm/sleefsimddp.c
```diff
@@ -268,7 +277,7 @@ EXPORT CONST vdouble xsin(vdouble d) {
   d = vmla_vd_vd_vd_vd(dqh, vcast_vd_d(-PI_C), d);
   d = vmla_vd_vd_vd_vd(dql, vcast_vd_d(-PI_C), d);
   d = vmla_vd_vd_vd_vd(vadd_vd_vd_vd(dqh, dql), vcast_vd_d(-PI_D), d);
-  
+
```
fpetrogalli-arm: Please don't change this line, we don't have a coding standard that enforces to remove lines with spaces.

src/libm/sleefsimddp.c
```diff
@@ -289,7 +298,7 @@ EXPORT CONST vdouble xsin(vdouble d) {
 					vor_vo_vo_vo(visnegzero_vo_vd(r),
 						     vgt_vo_vd_vd(vabs_vd_vd(r), vcast_vd_d(TRIGRANGEMAX)))),
 		       vcast_vd_d(-0.0), u);
-  
+
```
fpetrogalli-arm: Please don't change this line, we don't have a coding standard that enforces to remove lines with spaces.

src/libm/mkrename.c
```diff
@@ -61,10 +68,10 @@ int main(int argc, char **argv) {
   else {
     char *wdp = argv[1];
     char *wsp = argv[2];
-    char *vdoublename = argv[3];
-    char *vfloatname = argv[4];
-    char *vintname = argv[5];
-    char *vint2name = argv[6];
+    char *vdoublename = argv[3], *vdoublename_escspace = escapeSpace(vdoublename);
```
fpetrogalli-arm: Can you do it the other way around, so that you don't have to change the variable used in the printf calls? We have a patch downstream that adds the AArch64 Vector PCS attribute [1] to the renaming functions (of course, we will upstream it), your changes will you be easier to merge if you leave the original name in the printf calls.

[1] https://developer.arm.com/products/software-development-tools/hpc/arm-compiler-for-hpc/vector-function-abi

Configure.cmake
```diff
@@ -422,3 +450,8 @@ endif()
 if (COMPILER_SUPPORTS_WEAK_ALIASES)
   set(COMMON_TARGET_DEFINITIONS ${COMMON_TARGET_DEFINITIONS} ENABLE_ALIAS=1)
 endif()
+
+# When cross compiling for ppc64, this bug-workaround is needed
+if(CMAKE_CROSSCOMPILING AND CMAKE_SYSTEM_PROCESSOR MATCHES "^(powerpc|ppc)64")
+  set(COMMON_TARGET_DEFINITIONS ${COMMON_TARGET_DEFINITIONS} UNDEF_USE_EXTERN_INLINES=1)
```
fpetrogalli-arm: I have commented all the places where you use this define, asking to add a comment saying that the use of `UNDEF_USE_EXTERN_INLINES` is platform specific.

I was thinking, instead of adding the comments, how about making the variable names specific to PPC64, so there is not doubt it target specific? May I ask you to change it to `POWER64_UNDEF_USE_EXTERN_INLINES`? With this name change, I am happy for you to avoid adding the comments all around.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - 3.5-ULP versions of sinh, cosh, tanh, sinhf, coshf, tanhf, and the
   corresponding testing functionalities are added.
   https://github.com/shibatch/sleef/pull/192
+- Power VSX target support is added to libm.
```
fpetrogalli-arm: Love this! :)

src/libm/mkrename.c
shibatch: Not all the arguments of printf calls are space-escaped.
I don't think of a good way of fulfilling your request.

src/dft/dft.c
```diff
@@ -291,7 +298,12 @@ static void dispatch(SleefDFT *p, const int N, real *d, const real *s, const int
 
 // Transposer
 
+#if defined(__GNUC__) && __GNUC__ < 5
```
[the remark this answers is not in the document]
shibatch: This is another bug workaround.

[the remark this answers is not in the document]
fpetrogalli-arm: Sorry, I meant to describe the bug workaround in the code, with a comment.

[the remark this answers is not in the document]
shibatch: It's also described in the code.

src/libm/mkrename.c
shibatch: In altivec or VSX intrinsics, vector data types are like "vector float".

https://gcc.gnu.org/onlinedocs/gcc/PowerPC-AltiVec_002fVSX-Built-in-Functions.html
http://openpowerfoundation.org/wp-content/uploads/resources/leabi/content/dbdoclet.50655245_56497.html

fpetrogalli-arm: h
Hi @shibatch , sorry for bothering you again with this. I had a second (or maybe third!!) thought, and if my understanding is correctly, the following variables don't have space in their names, unless when targeting VSX:
```
vdoublename
vfloatname
vintname
vint2name
```
 
Given that the function `escapeSpace` do not modify the values of those variables when not targeting VSX, I'd rather have you redefine them as follows:

```
char *vdoublename = escapeSpace(argv[3]);
char *vfloatname = escapeSpace(argv[4]);
char *vintname = escapeSpace(argv[5]);
char *vint2name = escapeSpace(argv[6]);
```

This should keep the functionality of the current patch, without the need to modify the variables name in the printf calls.

src/libm/mkrename.c
```diff
@@ -10,6 +10,13 @@
 
 #include "funcproto.h"
 
+char *escapeSpace(char *str) {
+  char *ret = malloc(strlen(str) + 10);
```
fpetrogalli-arm: This memory allocation is never released. you should modify the content of `str` in place, and definitely not return `ret`.

shibatch:
The contents of vdoublename and other variables are used for its own type name and part of typedef name. So, we need both the original one and the replaced one. They cannot be unified.

src/libm/mkrename.c
shibatch: I think this is harmless, but I added free for these variables.

fpetrogalli-arm:
> The contents of vdoublename and other variables are used for its own type name and part of typedef name. So, we need both the original one and the replaced one. They cannot be unified.

@shibatch Sigh - apologies, now I have seen it. Sorry for keeping you on this.

fpetrogalli-arm:
Thank you for adding this new target @shibatch! This shows all the goodness of SLEEF!

PR #204, opened 2018-07-19
Determinism

shibatch:
This patch adds functions that return consistent values across all platforms for deterministic simulation.
The functions whose names begin with Sleef_cinz are the functions that can be used on every platform.
The functions whose names begin with Sleef_finz are the functions that can be used only on platforms with FMA support.
Scalar functions are named like Sleef_cinz_sind1_u35purec.
On AArch32, only scalar functions are supported.
Please refer to the following PRs for detailed explanation of changes.

https://github.com/shibatch/sleef/pull/200 
https://github.com/shibatch/sleef/pull/201
https://github.com/shibatch/sleef/pull/202
https://github.com/shibatch/sleef/pull/203


shibatch:
@d-parks Please all give me comments about this PR.
This is basically what you requested in issue https://github.com/shibatch/sleef/issues/47, though the implementation is a little bit different from what I was planning.

d-parks:
Dear Shibata-san

Thank you. I'm traveling this week but I should be able to find some time
to download your latest commit.

Best regards

Dave

fpetrogalli-arm:
Hello @shibatch , thank you for working on this. I have to admit it is too big for me to be able to give a constructive review.

Would it be possible for you to re-open PRs #200, #201, #202 and #203 and request reviews for those against a new `Determinism` branch? By doing so, it will be easier for me to break down the review.

We can incrementally merge the four PRs into a new branch, and when we are happy with it, just merge the branch itself on master. 

Apologies for asking you to do some extra work.

Francesco

shibatch:
@fpetrogalli-arm I can do that if you will review each PR in a few days.
Please tell me what you are actually checking at each PR.
Then I will explain those points when I make a PR.

xoofx:
@shibatch starting to implement deterministic calculation, but I'm failing to find a function like `Sleef_cinz_cosf_u35` or a `Sleef_cinz_cosf4_u35sse4` for example. I missed likely something here... 😅

shibatch:
Sleef_cinz_cosf4_u35sse4 should be included in the library.
Sleef_cinz_cosf_u35 does not exists. It is Sleef_cinz_cosf1_u35purec.
Please use nm -g command to show the exported symbols.

shibatch:
Please use the source code from the following PR.
https://github.com/shibatch/sleef/pull/229

xoofx:
I was relying on this PR, but actually, that might come from the LLVM IR output that doesn't contain them (I'm not using compiled binaries). So It is likely on my side.

xoofx:
Ok got them running. I had to adjust the LLVM compilation. I will try to push a PR for that.

I have one question though: I can see that there are functions like `Sleef_cinz_sincospif4_u05sse4` which are not in the deterministic mode but generated in the plain SSE4 mode. I was expecting to get only deterministic functions if I setup `DETERMINISTIC=1` for the relevant bitcode LLVM files, but that doesn't seem the case.

shibatch:
>I have one question though: I can see that there are functions like Sleef_cinz_sincospif4_u05sse4 which are not in the deterministic mode but generated in the plain SSE4 mode.

I don't understand what you mean. What specific symbols do you expect to be included? At least Sleef_cinz_sincospif4_u05sse4 is included.

>I was expecting to get only deterministic functions if I setup DETERMINISTIC=1 for the relevant bitcode LLVM files, but that doesn't seem the case.

Those deterministic functions are also important for users other than you.
I haven't checked the generated bitcode, but is it important for you to make deterministic functions only included only if the option is specified? If so, I will try to make the library built in that way.

xoofx:
> I don't understand what you mean. What specific symbols do you expect to be included? At least Sleef_cinz_sincospif4_u05sse4 is included.
> Those deterministic functions are also important for users other than you.
I haven't checked the generated bitcode, but is it important for you to make deterministic functions only included only if the option is specified? If so, I will try to make the library built in that way.

I want all functions generated in the end, so I don't have a huge problem with that. It's just that part of the deterministic functions are generated when you don't specify DETERMINISTIC=1, so that's why I was confused. I thought it was a build problem on my side. The files generated when setting `DETERMINISTIC=1` were also much smaller (likely because the other part is in standard files), so I was not sure if I got the build right. If the union of the 2 gives me the correct result, fair enough, but that's not super intuitive.

shibatch:
Currently, DETERMINISTIC is not a build option. It is a flag that is automatically set when the library is built. The library should include all the deterministic version of functions with the default setting.

xoofx:
> Currently, DETERMINISTIC is not a build option. It is a flag that is automatically set when the library is built. The library should include all the deterministic version of functions with the default setting.

Ok, fair enough. When building for LLVM, you don't have access to these settings (some work would be required to copy them automatically), so that's why I need to replicate them manually. It's working with these [changes](https://github.com/xoofx/sleef/commit/666dfade17b98f0eac2d4fe9d17e8e867c49ab92), I will try to upstream them.

xoofx:
> Sounds good, but please remember that PR #229 is still open.
> Please open a new PR after approving #229.

Sure, I will be able to test #229 as part of determinism tests.

PR #205, opened 2018-07-20
Subset of changes required to make sleef compile on FreeBSD.

iotamudelta:
These are the changes for amd64 FreeBSD to make it compile that are not clear particularities of FreeBSD itself. These particularities are:
* isinff() is missing
* we need to typically add /usr/local/include to include_directories

Note that this does *not* imply sleef to work on FreeBSD - I haven't figured out the correct way to convince clang to recognize functions and vectorize properly. Ideas here are welcome!


shibatch:
For isinff(), there are alternative definitions in src/common/misc.h. Please try enabling them on FreeBSD.
For include_directories, please modify CMakeLists.txt. I think this is easy.

shibatch:
> I haven't figured out the correct way to convince clang to recognize functions and vectorize properly.

Does gcc work correctly on FreeBSD?
Could you paste the error generated by clang?

iotamudelta:
@shibatch I have the hack to `CMakeLists.txt` - this will be maintained as a patch in the port ultimately. I tried the hack - that was not enough and again have a local hack to use `__issinff()` instead, which we have. Ultimately I think we want add the alias in FreeBSD. So no worries here for you!

`gcc` works in principle but is not the default compiler any longer fro tier-1 architectures, `clang` is and hence my interest is to get that to work. If I try the factorial test from here, I get:

```
$ clang-devel -O3 -Rpass=loop-vectorize -Rpass-analysis=loop-vectorize -mavx2  -fno-honor-nans -fno-math-errno -fno-trapping-math /home/jmd/software/sleef/build_llvm/lib/libsleef.so /home/jmd/software/sleef/build_llvm/lib/libsleefgnuabi.so factorial.c -o factorial_clang
factorial.c:22:3: remark: the cost-model indicates that interleaving is not beneficial [-Rpass-analysis=loop-vectorize]
  for (int i = 0; i < N; i++) {
  ^
factorial.c:22:3: remark: vectorized loop (vectorization width: 4, interleaved count: 1) [-Rpass=loop-vectorize]
factorial.c:29:14: remark: loop not vectorized: call instruction cannot be vectorized [-Rpass-analysis=loop-vectorize]
    in[i] = (rand() / (double)RAND_MAX) * 10;
             ^
factorial.c:28:3: remark: loop not vectorized: read with atomic ordering or volatile read [-Rpass-analysis=loop-vectorize]
  for(int i=0;i<N;i++)
  ^
factorial.c:22:3: remark: the cost-model indicates that interleaving is not beneficial [-Rpass-analysis=loop-vectorize]
  for (int i = 0; i < N; i++) {
  ^
factorial.c:22:3: remark: vectorized loop (vectorization width: 4, interleaved count: 1) [-Rpass=loop-vectorize]
factorial.c:34:38: remark: loop not vectorized: call instruction cannot be vectorized [-Rpass-analysis=loop-vectorize]
    printf("%.20g, %.20g\n", out[i], gamma(in[i]+1));
                                     ^
factorial.c:33:3: remark: loop not vectorized: read with atomic ordering or volatile read [-Rpass-analysis=loop-vectorize]
  for(int i=0;i<N;i++)
  ^
/usr/bin/ld: error: undefined symbol: pow
>>> referenced by factorial.c
>>>               /tmp/factorial-1eb3fb.o:(factorialN)

/usr/bin/ld: error: undefined symbol: pow
>>> referenced by factorial.c
>>>               /tmp/factorial-1eb3fb.o:(factorialN)

/usr/bin/ld: error: undefined symbol: pow
>>> referenced by factorial.c
>>>               /tmp/factorial-1eb3fb.o:(factorialN)

/usr/bin/ld: error: undefined symbol: pow
>>> referenced by factorial.c
>>>               /tmp/factorial-1eb3fb.o:(factorialN)

/usr/bin/ld: error: undefined symbol: pow
>>> referenced by factorial.c
>>>               /tmp/factorial-1eb3fb.o:(factorialN)

/usr/bin/ld: error: undefined symbol: pow
>>> referenced by factorial.c
>>>               /tmp/factorial-1eb3fb.o:(factorialN)
....
```

shibatch:
I now understand what you are trying to do.
Unfortunately, that is not possible at this time. They are still working on it.
The only way you can utilize SLEEF is to manually vectorize your code and call functions like Sleef_sind2_u10.

Issue #206, opened 2018-08-13
Make it easier to auto-generate bindings for the library


gnzlbg:
Is there a configuration option I can use to disable all `long double`-related functionality ? 

I am trying with `-DCOMPILER_SUPPORTS_LONG_DOUBLE=FALSE` and `-DCOMPILER_SUPPORTS_FLOAT128=FALSE` but I do still get types like `Sleef_longdouble2` in the generated headers.

I do not mind about the library using `long double` internally, but about including it in the exported headers.

gnzlbg:
I am auto-generating sleef bindings for Rust, which does not have a native 80-bit floating point type. 

Not being able to hide the APIs using 80-bit floats makes this task much harder.

I would also like to expose the APIs compiled with different features e.g. I'd like to expose sse2, sse4.2, avx, avx2, and avx-512 versions of `sinf_u1`, but it is unclear to me whether all these versions are currently compiled into the library and their symbols exposed, or whether I have to manually build the library N times for each set of features, merge them somehow, and the wrap that.

shibatch:
>Not being able to hide the APIs using 80-bit floats makes this task much harder.

I don't fully understand what you are trying to do, but simply removing unnecessary lines using grep is not enough?

> it is unclear to me whether all these versions are currently compiled into the library and their symbols exposed, or whether I have to manually build the library N times for each set of features, merge them somehow, and the wrap that.

All the features that can be compiled by the C compiler are compiled in. It does not make sense to build the library multiple times.

fpetrogalli-arm:
> I would also like to expose the APIs compiled with different features e.g. I'd like to expose sse2, sse4.2, avx, avx2, and avx-512 versions of sinf_u1, but it is unclear to me whether all these versions are currently compiled into the library and their symbols exposed, or whether I have to manually build the library N times for each set of features, merge them somehow, and the wrap that.
> 
Hi @gnzlbg,

You have two choices here: you can either call the functions from libsleef, which are mangled according to the scheme in http://sleef.org/x86.xhtml (figure 3.1), or you can use the library libsleefgnuabi which is libmvec (vector math library in glib) and that uses the naming conventions and ABI defined in the Vector Function ABI for x86 [1]. AArch64 has a similar ABI too [2]. The two libraries provide the same list of vector functions.

Francesco

[1] https://sourceware.org/glibc/wiki/libmvec?action=AttachFile&do=view&target=VectorABI.txt
[...]

PR #208, opened 2018-08-15
Add options to disable and enforce features

shibatch:
This patch adds CMake options for disabling and enforcing detected features.

@gnzlbg Please check if this patch implements all what you want.

https://github.com/shibatch/sleef/issues/206


gnzlbg:
This is even better than what I wanted since I can also use it to disable some other features that Rust doesn't fully support yet like FMA4 and AVX-512F. So thanks!

PR #213, opened 2018-08-16
[Determinism] Add purecscalar and noFMA helpers

shibatch:
This patch adds pure C scalar helpers with which scalar functions are implemented using the SIMD source code.
It also adds helper files which do not use FMA on platforms with FMA support.

### DIFF
--- a/travis/before_install.x86_64-gcc.sh
+++ b/travis/before_install.x86_64-gcc.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 set -ev
-#tar xf sde-external-8.12.0-2017-10-23-lin.tar.bz2
 sudo add-apt-repository -y ppa:adrozdoff/cmake
 sudo apt-get -qq update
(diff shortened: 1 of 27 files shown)


travis/before_install.x86_64-gcc.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -ev
-tar xf sde-external-8.12.0-2017-10-23-lin.tar.bz2
+#tar xf sde-external-8.12.0-2017-10-23-lin.tar.bz2
```
fpetrogalli-arm: Why is this commented out? Is it because the AVX512 tests are run on the jenkins machine? If so, you should remove the line, and possibly also the bits in the travis configuration that use that variable. But in fact, I would like to see this remove of SDE from Travis in a separate PR/patchset, as it is unrelated to adding the purec_scalar targets.

shibatch: I now removed all SDE-related lines

PR #216, opened 2018-08-17
[Determinism] Add deterministic functions

shibatch:
This patch adds implementations of deterministic functions.

The SIMD source files(sleefsimd?p.c) are compiled twice for each vector extension, with DETERMINISTIC macro turned on and off.
Renaming by rename*.h is switched according to DETERMINISTIC macro.

When DETERMINISTIC macro is undefined, the function name xsin will be renamed to Sleef_sind2_u35sse2 with renamesse2.h, for example.

If DETERMINISTIC macro is defined, the function name xsin will be renamed to Sleef_cinz_sind2_u35sse2, for example.

iuty* and tester2y* are added in order to test the newly added deterministic functions. As a consequence, time for testing is increased to almost two times.


fpetrogalli-arm:
Are there any specific changes I should look for in `libsleefgnuabi`?

shibatch:
libsleefgnuabi should be unchanged.

fpetrogalli-arm:
Thank you. Does it mean that the deterministic versions of the functions are not compiled in libsleefgnuabi?

shibatch:
At this time, no new functions are added to libsleefgnuabi.
If you want them to be added, please write an issue explaining how they are named.

src/libm/sleefsimddp.c
```diff
@@ -389,6 +393,13 @@ static INLINE CONST ddi_t rempi(vdouble a) {
   return ret;
 }
 
+#if !defined(DETERMINISTIC)
```
fpetrogalli-arm: Given that the function is always called "xsin" in C code, I would prefer to have the `#if !defined(DETERMINISTIC)` switch _inside_ the function definition:


```
EXPORT CONST vdouble xsin(vdouble d) {
#if !defined(DETERMINISTIC)
// original version
#else // #if !defined(DETERMINISTIC)
// deterministic version
#endif // #if !defined(DETERMINISTIC)
}
```

Could you please apply this change to all the function definitions of this change-set in which you have introduced DETERMINISTIC behavior?

This is quite important because our downstream version of SLEEF is using the attribute `aarch64_vector_pcs` [1],  and having only one declaration instead of two will make merging easier.

[1] https://developer.arm.com/products/software-development-tools/hpc/arm-compiler-for-hpc/vector-function-abi section 2.1

fpetrogalli-arm:
> At this time, no new functions are added to libsleefgnuabi.
> If you want them to be added, please write an issue explaining how they are named.

Thank you for explaining, I don't see a need to add deterministic functions to `libsleefgnuabi` for now, we can come back to it in the future if needed. I just wanted to make sure that we could keep using `libsleefgnuabi` in our compiler after the changes in this patch.

shibatch:
>I don't understand how the DETERMINISTIC version are tested with CI in this patch. Could you please explain how this is happening? Or point at CI logs showing that the deterministic tests are executed together with the non deterministic ones?

The iut programs whose name begins with "iuty" are the iut for the deterministic version of functions. By checking the result of testing with iutysse2, for example, you can be sure that the corresponding deterministic functions passes the accuracy and nonnumber tests. I will add bit-identity tests in the next patch.

src/libm-tester/CMakeLists.txt
```diff
@@ -85,6 +85,20 @@ macro(test_extension SIMD)
     add_test_iut(${TARGET_IUT${SIMD}})
     list(APPEND IUT_LIST ${TARGET_IUT${SIMD}})
 
+    string(CONCAT IUTYNAME "iuty" ${LCSIMD})
```
fpetrogalli-arm: Please add a comment in the source explaining the difference between iut and iuty

src/libm-tester/CMakeLists.txt
```diff
@@ -100,6 +114,18 @@ macro(test_extension SIMD)
 	      if (MPFR_INCLUDE_DIR)
 		target_include_directories(${T} PRIVATE ${MPFR_INCLUDE_DIR})
 	      endif()
+
+	      set(T "tester2y${SCSIMD}${P}")
```
fpetrogalli-arm: Same here, make sure the distinction between the old tester2 and tester2y is clear

src/libm/sleefsimdsp.c
```diff
@@ -2472,6 +2965,77 @@ EXPORT CONST vfloat xerfcf_u15(vfloat a) {
   r = vsel_vf_vo_vf_vf(visnan_vo_vf(s), vcast_vf_f(SLEEF_NANf), r);
   return r;
 }
+#endif // #if !defined(DETERMINISTIC)
+
+#if !defined(DETERMINISTIC) && !defined(ENABLE_GNUABI)
+
+#ifdef ENABLE_ALIAS
```
fpetrogalli-arm: Can you explain what are these alias for? Please do it in a comment in the code.

shibatch:
@xoofx No problem.
It must be a good thing for the industry to have paradigm shifts frequently. 😃

PR #219, opened 2018-08-19
Fix i386 issues

shibatch:
This patch fixes build problem on i386(32-bit) linux systems.
Testing on Jenkins is added.
expfrexpf and ilogbf are removed from gnuabi_compatibility.c, since these two functions are not officially supported and cause a build problem.


src/libm-tester/gnuabi_compatibility.c
```diff
@@ -222,7 +221,6 @@ DECLARE_SP(fminf, vv);
 DECLARE_SP(fmodf, vv);
 DECLARE_SP(frfrexpf, v);
 DECLARE_SP(hypotf, vv);
-DECLARE_SP(ilogbf, v);
```
fpetrogalli-arm: Please don't do this, we use `ilogbf` in arm compiler for HPC. Is it possible to fix the build error? What do you mean with "it is not officially supported"?

shibatch: The problem is that the functions are called with different types of arguments from the API in gnuabi_compatiblity.c. I cannot fully explain, but the program segfaults on i386 with these two functions.

At this time, AVX version uses a data type different from AVX2 version for vint2. This is because AVX does not support 256-bit integer operations. Because of this, the functions that expose vint2 are not included in the officially supported functions.

fpetrogalli-arm: The API in `gnuabi_compatibility.c` is a hack that is there just to simplify the number of macros that we are using in the sources. If the problem is an AVX-only problem, then I think that `#ifdef` guard the call in the sources should be enough. Of course, this breaks a bit the claim of SLEEF being target independent, as we actually are saying that one function is not working for AVX - am I right?

shibatch: At this time, we are not claiming that those two functions work.
Anyway, I will add #ifdef to the two functions.

src/libm-tester/gnuabi_compatibility.c
```diff
@@ -222,7 +221,10 @@ DECLARE_SP(fminf, vv);
 DECLARE_SP(fmodf, vv);
 DECLARE_SP(frfrexpf, v);
 DECLARE_SP(hypotf, vv);
+#ifndef ENABLE_AVX
```
fpetrogalli-arm: it would be good to track this issue in github, and mention it here (and below in the other IFDEF).

shibatch: I think the a proper way to resolve this issue is to call the functions with the correct types of arguments.

fpetrogalli-arm: I see, OK. Then, an issue is needed to track the problem. Please also add how you think this can be solved.

shibatch: And it does not need to be resolved.
Another way is to just remove those two functions.

Issue #217, opened 2018-08-18
`double` copysignfk and xisnumberf in sleefsp


burrbull:
https://github.com/shibatch/sleef/blob/d22877e4c6b6bc20fa9548ec2cfff36aee3cdbcf/src/libm/sleefsp.c#L59

Is this need to be double?

shibatch:
This is a bug.
Thank you for pointing out.

shibatch:
This is now fixed.

PR #220, opened 2018-08-20
Add FreeBSD support

shibatch:
This patch adds support for FreeBSD.
Testing on Jenkins is added.
The bug pointed out on issue https://github.com/shibatch/sleef/issues/217 is also fixed.


Configure.cmake
```diff
@@ -608,10 +608,23 @@ CHECK_C_SOURCE_COMPILES("
   }"
   COMPILER_SUPPORTS_SYS_GETRANDOM)
 
+# Check if compilation with FFTW3 really succeeds
```
fpetrogalli-arm: Hum - this seems wrong to me. What is the reason for libfftw to fail compiling?

src/common/misc.h
```diff
@@ -285,4 +285,11 @@ static INLINE CONST int isnanf(float x) { return x != x; }
 static INLINE CONST int isnanl(long double x) { return x != x; }
 #endif
 
+#if defined(__FreeBSD__)
```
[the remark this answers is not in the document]
shibatch: inff and infl are declared in the header but not defined in libm on FreeBSD.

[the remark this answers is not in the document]
fpetrogalli-arm: So, why not use the macro here for all OSs? This will reduce the amount of code duplication.

src/common/misc.h
```diff
@@ -271,18 +271,13 @@ typedef struct {
...
-static INLINE CONST int isnanl(long double x) { return x != x; }
-
 #endif // defined(_MSC_VER)
 
-#if defined(__APPLE__) || defined(__MINGW32__) || defined(__MINGW64__)
-static INLINE CONST int isinff(float x) { return x == SLEEF_INFINITYf || x == -SLEEF_INFINITYf; }
-static INLINE CONST int isinfl(long double x) { return x == SLEEF_INFINITYl || x == -SLEEF_INFINITYl; }
-static INLINE CONST int isnanf(float x) { return x != x; }
-static INLINE CONST int isnanl(long double x) { return x != x; }
+#if defined(__APPLE__) || defined(__MINGW32__) || defined(__MINGW64__) || defined(__FreeBSD__) || defined(_MSC_VER)
```
fpetrogalli-arm: This look better now. One more question: can we define these macro unconditionally? Or will it break the build on linux because those are not defined? If linux is the only platform where these are already available, can we turn the `#if defined(...) || ... || defined(..)` into a `#ifndef __linux__`?

src/libm-tester/tester2sp.c
```diff
@@ -38,6 +38,10 @@
 static int isinff(float x) { return x == __builtin_inff() || x == -__builtin_inff(); }
 #endif
 
+#if defined(__FreeBSD__) 
+#define isinff(x) ((x) == (float)(1e+300) || (x) == -(float)(1e+300))
```
fpetrogalli-arm: this is already defined in /src/misc/common.h and it is different:

```#define isinff(x) ((x) == SLEEF_INFINITYf || (x) == -SLEEF_INFINITYf)```

Which is the correct one for freebsd?

shibatch: Both are correct.
SLEEF_INFINITYf is not defined in sleefsp.c, and so (float)(1e+300) is used instead.

src/common/misc.h
shibatch: I think it is not clean to define those macros unconditionally since they are semi-standardized.
I don't know if these macros are only defined in linux systems.

PR #223, opened 2018-08-29
[Tester] Faster tester and iut

shibatch:
With this patch, testing time is reduced to 50%.

### DIFF
--- a/src/libm-tester/tester.c
+++ b/src/libm-tester/tester.c
@@ -3371,6 +3371,7 @@ void do_test() {
   //
 
   if (enableDP) {
+    // 64 > 53(=number of bits in DP mantissa)
     mpfr_set_default_prec(64);
 
     fprintf(stderr, "hypot_u35 : ");
@@ -3568,6 +3569,7 @@ void do_test() {
 
     //
 
+    // 1280 > 1024(=maximum DP exponent) + 53(=number of bits in DP mantissa)
     mpfr_set_default_prec(1280);
 
     fprintf(stderr, "sin in sincospi_u35 : ");
@@ -4119,6 +4121,7 @@ void do_test() {
   //
 
   if (enableSP) {
+    // 53 > 24(=number of bits in SP mantissa)
     mpfr_set_default_prec(53);
 
     fprintf(stderr, "hypotf_u35 : ");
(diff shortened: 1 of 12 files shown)


src/libm-tester/tester.c
```diff
@@ -4311,7 +4316,7 @@ void do_test() {
 
     //
 
-    mpfr_set_default_prec(1280);
+    mpfr_set_default_prec(320);
```
fpetrogalli-arm: Did you reduce the precision of MPFR because it gives speedup? Why did you choose the original values? Are these new values enough to make sure we don't miss errors?

src/libm-tester/testerutil.c
```diff
@@ -124,7 +124,11 @@ int readln(int fd, char *buf, int cnt) {
 }
 
 int startsWith(char *str, char *prefix) {
-  return strncmp(str, prefix, strlen(prefix)) == 0;
+  while(*str != '\0' && *prefix != '\0') {
```
fpetrogalli-arm: Are you sure this code is faster then the library call? I believe that strncmp have highly specialized implementations in glibc...

src/libm-tester/tester.c
```diff
@@ -47,6 +47,7 @@ void stop(char *mes) {
 
 int ptoc[2], ctop[2];
 int pid;
+FILE *fpctop;
```
fpetrogalli-arm: Does it have to be global? Can't it be private inside of main?

src/libm-tester/tester.c
shibatch: 1280 is for double precision.
320 is more than 128(exponent of single precision) + 24(mantissa length). Actually, 160 should be enough.

src/libm-tester/testerutil.c
shibatch: The original code is strncmp + strlen.
This code is faster if defined as a static function.
I will move this to testerutil.h.

src/libm-tester/tester.c
shibatch: fpctop is reference by many functions.

travis/script.aarch64-gcc.sh
```diff
@@ -2,6 +2,7 @@
 set -ev
 cd /build/build-cross
 make -j 2 all
+export OMP_WAIT_POLICY=passive
```
fpetrogalli-arm: What is this for?

shibatch: This is for changing OpenMP setting.
With the default setting, it does not run fast if there are many OpenMP processes running simultaneously.

fpetrogalli-arm:
@shibatch , where do you see the 50% speedup you claim in the PR?
When I look at https://travis-ci.org/shibatch/sleef/pull_requests, the run times seem comparable (~2h).

shibatch:
Running time at travis includes time for comiplation.
Compare, for example https://travis-ci.org/shibatch/sleef/jobs/422911913 and https://travis-ci.org/shibatch/sleef/jobs/418961450.
Testing time for iut is reduced from 165.04 sec to 89.70 sec.

PR #228, opened 2018-10-15
Merging 3.3.1 with aavpcs

fpetrogalli-arm:
This is a work in progress of the merge I am doing to import the AAVPCS (AArch64 Vector Procedure Call Standard) introduced by the Vector Function ABI specifications for AArch64.


fpetrogalli-arm:
@shibatch , our downstream version of SLEEF is based on 3.3.1, which is where I applied the AAVPCS changes originally. I thought it was a good idea to merge that 3.3.1-based branch onto master, but I picked up the commits of 3.3.1 that where not in master. Let me know if this is OK for for you, or if you want me to throw away those 3.3.1-only commits. 

By the way, how comes we have stuff that is only in 3.3.1 but not in master?

One final note: I see things failing, even locally (for example, tester3 is not happy on Advanced SIMD with the AAVPCS), but I create the pull request just to see how much I am breaking upstream in the other builders.

src/arch/helperadvsimd.h
```diff
@@ -90,137 +90,137 @@ static INLINE vfloat vgather_vf_p_vi2(const float *ptr, vint2 vi2) {
...
-static INLINE vint2 vtruncate_vi2_vf(vfloat vf) { return vcvtq_s32_f32(vf); }
-static INLINE vfloat vcast_vf_vi2(vint2 vi) { return vcvtq_f32_s32(vi); }
-static INLINE vint2 vcast_vi2_i(int i) { return vdupq_n_s32(i); }
-static INLINE vint2 vrint_vi2_vf(vfloat d) {
-  return vcvtq_s32_f32(vrndnq_f32(d));
+static INLINE VECTOR_CC vint2 vtruncate_vi2_vf(vfloat vf) { return vcvtq_s32_f32(vf); }
+static INLINE VECTOR_CC vfloat vcast_vf_vi2(vint2 vi) { return vcvtq_f32_s32(vi); }
+static INLINE VECTOR_CC vint2 vcast_vi2_i(int i) { return vdupq_n_s32(i); }
+static INLINE VECTOR_CC vint2 vrint_vi2_vf(vfloat d) {
+  return vcvtq_s32_f32(vrndaq_f32(d));
```
shibatch: This line should be 
```
return vcvtq_s32_f32(vrndnq_f32(d));
```

shibatch: This line should be 
```
return vcvtq_s32_f32(vrndnq_f32(d));
```

shibatch:
```
```suggestion
  return vcvtq_s32_f32(vrndaq_f32(d));
This line should be
```
return vcvtq_s32_f32(vrndnq_f32(d));
```

shibatch: Revert this line

fpetrogalli-arm:
@shibatch, thank you for fixing the numerical problems.

> Isn't it better to check whether AAVPCS is supported by the compiler with CHECK_C_SOURCE_COMPILES?

We thought about that when we implemented it downstream, but given that the library is used in both situations (AAVPCS and no AAVPCS) and not all compilers support the feature right now, I'd rather have it explicitly passed at configuration time.

Let me know if you are happy for me to submit this change.

PR #229, opened 2018-10-28
Add low accuracy sin, cos, log2, exp2 and pow

shibatch:
I added the following functions.
* fastsinf_u3500
* fastcosf_u3500
* fastpowf_u3500
* log2_u35, log2f_u35
* exp2_u35, exp2f_u35

The error bound of fastsin and fastcos are max(2e-6, 350ULPs).
Each function has corresponding deterministic functions.

@xoofx Please let me know how you think about these functions.


xoofx:
Amazing, will have a look, thanks!

shibatch:
Testing at Travis is failing because brew is trying to build gcc from its source code.

fpetrogalli-arm:
Why introduce the `fast` prefix in the names?

fpetrogalli-arm:
> Testing at Travis is failing because brew is trying to build gcc from its source code.

Is this something that is broken also on master? If so, I think we should fix master first...

shibatch:
I am hoping that this "problem" will be fixed at brew side.
Otherwise, we need to change the CI setting, of course.

PR #231, opened 2018-12-09
delete output that is never generated

zdevito:
As far as I can tell, mkalias never generates the first file in this output list, causing this rule to appear always out-of-date, which causes a bunch of dependencies downstream to always rebuild.


shibatch:
Hello,
Thank you for your bug fix.
Please wait a few days since I need to fix the CI configuration.

shibatch:
This fix is included in PR https://github.com/shibatch/sleef/pull/236 and it is now merged.
Thank you for your report.

Issue #233, opened 2019-01-13
Add a vectorized quadruple precision math library


shibatch:
I am planning to add a vectorized quadruple precision math library to SLEEF. This is basically to add a quadruple precision version of the math functions in SLEEF.

This library is added as a new subproject of SLEEF. The existing subprojects won't be affected much.

The format of the FP number will basically comform to IEEE 754 quadruple-precision binary FP format. It will support denormal numbers, NaN and Infinity.

The computation methods are backed by triple-double arithmetic(TD). The subroutines basically first converts the argument in IEEE format to TD, evaluates the function in TD, and converts the result back to IEEE format. TD is a lot faster than computing in IEEE format, since packing to/unpacking from IEEE format is tricky and time consuming.

The library also includes subroutines for basic arithmetic operations(i.e. add, mul, div, compare and cast). At this point, I only have those routines backed by TD, and thus they are not fast, nor perfectly correctly rounded(their error bound is like 0.5000000001 ULP). I need to implement subroutines with direct computation methods in order to make them faster and correctly rounded.

The library will define new data types like the following. When a 128-bit vector extension, this data structure is basically a concatenation of two QP floating-point number.

```
typedef struct {
  vmask x, y;
} Sleef_quad2;
```

With 128-bit vector, they compute two QP values at a time. With purec helper, they compute one QP value.

It will define functions like the following.

```
Sleef_quad2 Sleef_sqrtq2_u05sse2(Sleef_quad2 a);

Sleef_quad Sleef_addq1_u05purec(Sleef_quad a, Sleef_quad b);
Sleef_quad2 Sleef_addq2_u05sse2(Sleef_quad2 a, Sleef_quad2 b);

__m128i Sleef_cmpltq2_sse2(Sleef_quad2 a, Sleef_quad2 b);

__m128d Sleef_castdq2_sse2(Sleef_quad2 a);
Sleef_quad2 Sleef_castqd2_sse2(__m128d a);
```

It will also define utility functions for converting from/to a string.

```
Sleef_quad Sleef_strtoq(const char *nptr, char **endptr, int base);
void Sleef_qtostr(char *s, size_t n, Sleef_quad a, int base);
```

The subroutine will contain tester and tester2, which correspond to the existing ones.

fpetrogalli-arm:
@shibatch , can we consider this done after #235 ? Or do you plan to add more functions?

shibatch:
I think it is okay to close this issue.
[...]

Issue #232, opened 2018-12-19
AArch64: -march=armv8-a+simd is incorrect


steleman:
This is based on sleef 3.3.1:

The cmake file Configure.cmake contains the following:

```
set(CLANG_FLAGS_ENABLE_NEON32 "--target=arm-linux-gnueabihf;-mcpu=cortex-a8")
# Arm AArch64 vector extensions.
set(CLANG_FLAGS_ENABLE_ADVSIMD "-march=armv8-a+simd")
```

This -march option is incorrect and unnecessary. It causes conflicts with -mcpu.

For example, GCC correctly warns if -mcpu=thunderx2t99 -march=armv8-a+simd was passed on compile-line:

```
%> cat t.c
#include <stdio.h>

int main()
{
    return 0;
}
```

```
%> gcc --version
gcc (GCC) 8.2.0
Copyright (C) 2018 Free Software Foundation, Inc.
[ ... ]
%> gcc -O3 -std=c99 -mcpu=thunderx2t99 -march=armv8-a+simd t.c -o t
cc1: warning: switch -mcpu=armv8.1-a conflicts with -march=armv8-a switch
```

Furthermore, this encourages building sleef without passing -mcpu=[X] on compile-line. Without -mcpu, the compiler will always choose the default generic instruction scheduler, which will lead to very poor performance on some AArch64 CPUs. For example on T99.

-mcpu=[X] is sufficient and should be required in CFLAGS. Either the particular AArch64 micro-arch can do V8 SIMD, or it can't. If it can do V8 SIMD, it will be available by default, without the need to pass -march=armv8-a+simd on compile-line.

fpetrogalli-arm:
Fixed in https://github.com/shibatch/sleef/commit/8e6e52f2153168ea7f7611076b6f077ea63003ac. Thank you @shibatch .

PR #235, opened 2019-01-22
Add quadprecision math library

shibatch:
This is a part of implementation of issue https://github.com/shibatch/sleef/issues/233.
At this point, add, mul, div and sqrt are implemented. Remaining functions will be committed in the succeeding PRs.
As for vector extensions, SSE2, AVX, FMA4, AVX2, AV2_128, AVX512F and AdvSIMD are supported.

This quadprecision math library is built only if -DBUILD_QUAD option is given to cmake. For some time(1 year?), this subproject is positioned at alpha development stage.

### DIFF
--- a/src/arch/helperadvsimd.h
+++ b/src/arch/helperadvsimd.h
@@ -703,11 +703,5 @@ static INLINE vopmask vgt64_vo_vm_vm(vmask x, vmask y) {
   return vreinterpretq_u32_u64(vcgtq_s64(vreinterpretq_s64_u32(x), vreinterpretq_s64_u32(y)));
 }
 
-static INLINE vmask vsll64_vm_vm_vm(vmask x, vmask c) {
-  return vreinterpretq_u32_u64(vshlq_u64(vreinterpretq_u64_u32(x), vreinterpretq_s64_u32(c)));
-}
-static INLINE vmask vsrl64_vm_vm_vm(vmask x, vmask c) {
-  return vreinterpretq_u32_u64(vshlq_u64(vreinterpretq_u64_u32(x), vnegq_s64(vreinterpretq_s64_u32(c))));
-}
 #define vsll64_vm_vm_i(x, c) vreinterpretq_u32_u64(vshlq_n_u64(vreinterpretq_u64_u32(x), c))
 #define vsrl64_vm_vm_i(x, c) vreinterpretq_u32_u64(vshrq_n_u64(vreinterpretq_u64_u32(x), c))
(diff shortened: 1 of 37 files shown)


shibatch:
I forgot to mention that "-march=armv8-a+simd" option is removed as pointed out in issue https://github.com/shibatch/sleef/issues/232.
The work around for ICE of armclang is also removed.

src/quad-tester/tester2simdqp.c
```diff
@@ -0,0 +1,326 @@
...
+  vargquad a0, a1, a2, a3;
+  Sleef_quad q0, q1, q2, q3, t;
+  mpfr_t frw, frx, fry, frz;
+  mpfr_inits(frw, frx, fry, frz, NULL);
+
+  for(cnt = 0;ecnt < 1000;cnt++) {
+    int e = cnt % VECTLENDP;
+
+    switch(cnt & 127) {
+    case 127:
```
fpetrogalli-arm: The `case` values should be handle by enums, with description of the meaning, otherwise it will be difficult to know what is going on here.

src/quad/CMakeLists.txt
```diff
@@ -0,0 +1,152 @@
...
+
+    set_target_properties(${OBJECT} PROPERTIES ${COMMON_TARGET_PROPERTIES})
+    add_dependencies(${OBJECT} qrename${SIMD}.h_generated)
+    target_compile_options(${OBJECT} PRIVATE ${FLAGS_ENABLE_${SIMD}})
+
+    list(APPEND SLEEFQUAD_OBJECTS $<TARGET_OBJECTS:${OBJECT}>)
+  endif()
+endforeach()
+
+add_library(sleefquad ${SLEEFQUAD_OBJECTS})
```
fpetrogalli-arm: Is there any reason why this is built as a separate library and not as part of libsleef?

shibatch:
>Is there any reason why the quad precision functions and tests live in a separate folder and not in src/libm?

There is so much difference in code maturity and public demand.
I think there is smaller number of people who are interested in the qp library.
I am going to add many qp functions in coming years.
If we keep qp functions in the same folder as libm, we will have much more hustle.

src/arch/helperadvsimd.h
```diff
@@ -644,3 +648,60 @@ static INLINE VECTOR_CC void vsscatter2_v_p_i_i_vf(float *ptr, int offset, int s
   vst1_f32((float *)(ptr+(offset + step * 0)*2), vget_low_f32(v));
   vst1_f32((float *)(ptr+(offset + step * 1)*2), vget_high_f32(v));
 }
+
```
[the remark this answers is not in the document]
shibatch: SVE version is coming.
Because of variable vector length, I need to check more things on SVE than other vector extensions.

PR #237, opened 2019-01-27
[Quad] Add functions for conversion between quad and string

shibatch:
This patch adds Sleef_strtoq and Sleef_qtostr which can be used to convert between a quad value and a string. These functions are not vectorized. The corresponding testers are also added.

This patch also adds functions for subtraction.

Intel compiler testing is temporarily disabled because of license expiration.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -43,7 +43,7 @@ pipeline {
 			 '''
             	     }
                 }
-
+/*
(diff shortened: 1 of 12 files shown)


fpetrogalli-arm:
Please create an issue to track reactivating the Intel Compiler CI. Then, feel free to squash this PR (not merge!).

PR #239, opened 2019-01-29
[LIBM] Introduce faster method for evaluating polynomials

shibatch:
This patch replaces Horner method which was used to evaluate polynomials with Estrin's method( https://en.wikipedia.org/wiki/Estrin%27s_scheme ) that allows more parallel computations with out-of-order execution.
This patch also introducing a new reduction method to tan.
With this patch, mainly computation for double-precision functions becomes faster, and the effect is like a few percent to 20 percent. For example, the ratio between execution time of the following functions before and after applying this patch is shown below.

Sleef_atan2d4_u35 : 1.21
Sleef_powd4_u10 : 1.17
Sleef_sind4_u35 : 1.10
Sleef_tand4_u10 : 1.04
Sleef_tand4_u35 : 1.17


fpetrogalli-arm:
Nice stuff. Could you please provide a speedup report on both Intel and AArch64?

shibatch:
Below is the graphs with x86.
x86 trig
x86 non-trig

shibatch:
Below is the graphs with aarch64.

aarch64 trig
aarch64 non-trig

src/libm/sleefdp.c
```diff
@@ -30,6 +30,36 @@ extern const double rempitabdp[];
 #pragma fp_contract (off)
 #endif
 
+#define POLY2(x, c1, c0) mla(x, c1, c0)
```
fpetrogalli-arm: These macros seem to be the same in all `sleef*.c` files. I think it would make sense to merge them in a shared header file (`misc.h`?)

PR #241, opened 2019-02-25
Add missing VECTOR_CC for sincos

yuyichao:
Found this when trying to use clang to compile sleef on windows with vectorcall calling convention. The calling convention changes the symbol name making this very easy to detect when fixing aliases.

### DIFF
--- a/src/libm/sleefsimddp.c
+++ b/src/libm/sleefsimddp.c
@@ -955,7 +955,7 @@ EXPORT CONST VECTOR_CC vdouble xcos_u1(vdouble d) {
 #define XMODF xmodf
 #endif
 
-TYPE2_FUNCATR vdouble2 XSINCOS(vdouble d) {
+TYPE2_FUNCATR VECTOR_CC vdouble2 XSINCOS(vdouble d) {
 #if !defined(DETERMINISTIC)
   vopmask o;
   vdouble u, t, rx, ry, s;
@@ -1101,7 +1101,7 @@ TYPE2_FUNCATR vdouble2 XSINCOS(vdouble d) {
 #endif // #if !defined(DETERMINISTIC)
 }
 
-TYPE2_FUNCATR vdouble2 XSINCOS_U1(vdouble d) {
+TYPE2_FUNCATR VECTOR_CC vdouble2 XSINCOS_U1(vdouble d) {
 #if !defined(DETERMINISTIC)
   vopmask o;
   vdouble u, rx, ry;
@@ -1262,7 +1262,7 @@ TYPE2_FUNCATR vdouble2 XSINCOS_U1(vdouble d) {
 }
 
 #if !defined(DETERMINISTIC)
-TYPE2_FUNCATR vdouble2 XSINCOSPI_U05(vdouble d) {
+TYPE2_FUNCATR VECTOR_CC vdouble2 XSINCOSPI_U05(vdouble d) {
   vopmask o;
   vdouble u, s, t, rx, ry;
   vdouble2 r, x, s2;


shibatch:
Hello yuyichao,

Is it possible that you also commit the necessary change for compiling SLEEF with clang on Windows?

yuyichao:
The full change is here. https://github.com/yuyichao/sleef/tree/win-vec

It's definately not in a mergable state and I was just trying to get it compile in the past 5 hours..... I don't think I have a lot of time to clean this up but I might. In any case, feel free to use it as a reference of the change needed.

The PKGBUILD I used to compile in msys2 is https://github.com/nacs-lab/nacs-pkgs/tree/3ff998f4b62169629c4813ffcf3d57cdeec6c511/windows/mingw-w64-sleef-git

yuyichao:
I'll add that compiling with clang doesn't need much change. It seems to work more or less out of the box. (well, at least it compiles right out of the box) It's making it working with the vector ABI that is tricky.

I'm not sure what's the status of sleef on windows (e.g. which ABI you are using). MSVC supports vectorcall while gcc doesn't yet so I'm not sure if it is desired to switch ABI on windows (the vectorcall ABI is AFAICT not callable from gcc ATM)

shibatch:
@fpetrogalli-arm Do you know something about the vector ABI on MSVC? I am also interested in supporting aarch64 with MSVC.

yuyichao:
https://docs.microsoft.com/en-us/cpp/cpp/vectorcall?view=vs-2017
https://clang.llvm.org/docs/AttributeReference.html#id335

I think aarch64 with MSVC doesn't need to worry about this since according to the doc (also sort of confirmed with clang x-compiling behavior) this attribute is ignored everything other than x86/x64.

shibatch:
Since ARM is an official partner of the project, I have to be careful if we are going to claim that aarch64 with MSVC is officially supported.

fpetrogalli-arm:
Hi all - I don't understand what this conversation is about, sorry. The change of this pull request makes absolutely sense (adding `VECTOR_CC` to `sincos`), but I don't see how it is related to MSVC support.

> @fpetrogalli-arm Do you know something about the vector ABI on MSVC? I am also interested in supporting aarch64 with MSVC.

The Vector Function ABI specifications are not related to any compiler. A compiler can decide or not whether it should comply to the specs.

> I think aarch64 with MSVC doesn't need to worry about this since according to the doc (also sort of confirmed with clang x-compiling behavior) this attribute is ignored everything other than x86/x64.

The attributes linked at https://docs.microsoft.com/en-us/cpp/cpp/vectorcall?view=vs-2017 and 
https://clang.llvm.org/docs/AttributeReference.html#id335 are not the same as the one used in this codebase. The `VECTOR_CC` macro in SLEEF resolves to `#define VECTOR_CC __attribute__((aarch64_vector_pcs))` when targeting `aarch64` with the `AAVPCS` enabled (see https://github.com/shibatch/sleef/blob/master/src/common/misc.h#L330). `AAVPCS` stands for Vector Procedure Call Standard, and it is described in the Vector Function ABI specifications at https://developer.arm.com/products/software-development-tools/hpc/arm-compiler-for-hpc/vector-function-abi

Please let me know if you have any other questions.

src/libm/sleefsimddp.c
```diff
@@ -955,7 +955,7 @@ EXPORT CONST VECTOR_CC vdouble xcos_u1(vdouble d) {
 #define XMODF xmodf
 #endif
 
-TYPE2_FUNCATR vdouble2 XSINCOS(vdouble d) {
+TYPE2_FUNCATR VECTOR_CC vdouble2 XSINCOS(vdouble d) {
```
fpetrogalli-arm: @yuyichao Please make sure that the declaration of `sincos` with this signature is also decorated with the `VECTOR_CC` in `sleef.h`.

yuyichao: I caught this because I get a linking error when the user code expected the calling convention from sleef.h but the definition code does not provide it so I believe it's already decorated in `sleef.h`. My understanding of `mkrename.c` also suggest that this should be the case.

fpetrogalli-arm: Yep, you are right, I verified on master, the VECTOR_CC is specified  for all versions of `sincos`:

```
frapet01@hw-a21-8:~/projects/sleef/build (master)$ cat include/sleef.h | grep sincos | grep aarch64
IMPORT CONST Sleef_float64x2_t_2 Sleef_sincosd2_u35advsimd(float64x2_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float64x2_t_2 Sleef_finz_sincosd2_u35advsimd(float64x2_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float64x2_t_2 Sleef_sincosd2_u10advsimd(float64x2_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float64x2_t_2 Sleef_finz_sincosd2_u10advsimd(float64x2_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float64x2_t_2 Sleef_sincospid2_u05advsimd(float64x2_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float64x2_t_2 Sleef_finz_sincospid2_u05advsimd(float64x2_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float64x2_t_2 Sleef_sincospid2_u35advsimd(float64x2_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float64x2_t_2 Sleef_finz_sincospid2_u35advsimd(float64x2_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float32x4_t_2 Sleef_sincosf4_u35advsimd(float32x4_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float32x4_t_2 Sleef_finz_sincosf4_u35advsimd(float32x4_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float32x4_t_2 Sleef_sincosf4_u10advsimd(float32x4_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float32x4_t_2 Sleef_finz_sincosf4_u10advsimd(float32x4_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float32x4_t_2 Sleef_sincospif4_u05advsimd(float32x4_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float32x4_t_2 Sleef_finz_sincospif4_u05advsimd(float32x4_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float32x4_t_2 Sleef_sincospif4_u35advsimd(float32x4_t) __attribute__((aarch64_vector_pcs));
IMPORT CONST Sleef_float32x4_t_2 Sleef_finz_sincospif4_u35advsimd(float32x4_t) __attribute__((aarch64_vector_pcs));
```

PR #242, opened 2019-02-25
Remove vectorcc from getInt and getPtr declarations

yuyichao:
These functions are not defined with vectorcc so they shouldn't be declared with it either.

Similar to https://github.com/shibatch/sleef/pull/241. The name change makes this very easy to catch...


yuyichao:
Done.

PR #243, opened 2019-02-25
Add (no-op) VECTOR_CC to sse/avx dispatch

yuyichao:
Currently a no-op but this makes it easier to support vector call attribute on x86,
which might be necessary on windows.

This is not a bug fix but it should also be harmless. This is likely the last part of the change I had to compile with clang+vectorcall ABI that is mergable. The [rest of the changes](https://github.com/shibatch/sleef/commit/5ccf30a25cbb98c91a2637ad4a12254b76021434) are fairly small (in line number) and much dirtier so it will certainly require cleanup and may or may not be wanted due to GCC compatibility... (since the two ABI uses different symbols, however, it might be possible to ship a binary with both version with the same function names....)

### DIFF
--- a/src/libm/dispavx.c.org
+++ b/src/libm/dispavx.c.org
@@ -66,76 +66,76 @@ static int cpuSupportsFMA4() {
 #endif
 
 #define DISPATCH_vf_vf(fptype, funcName, pfn, dfn, funcavx, funcfma4, funcavx2) \
-  static CONST fptype (*pfn)(fptype arg0);				\
-  static CONST fptype dfn(fptype arg0) {				\
-    fptype CONST (*p)(fptype arg0) = funcavx;				\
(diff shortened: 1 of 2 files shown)


yuyichao:
As I said, this is

> Currently a no-op but this makes it easier to support vector call attribute on x86,
> which might be necessary on windows.

Also see my link to

> The [rest of the changes](https://github.com/shibatch/sleef/commit/5ccf30a25cbb98c91a2637ad4a12254b76021434)

What other information do you need?

shibatch:
Although the rest of the changes is not included in this PR, I don't understand why you need to add a suffix like @@32.

yuyichao:
https://docs.microsoft.com/en-us/cpp/cpp/vectorcall?view=vs-2017

> Function names are suffixed with two "at" signs (@@) followed by the number of bytes (in decimal) in the parameter list.

PR #247, opened 2019-03-05
Add basic config install

xsacha:
Will allow you to find with find_package(sleef) and use with target_link_libraries(${TARGET_NAME} sleef::sleef).

Only basic right now as I am unfamilar with the project and not sure if there are any special cmake defines that need to be passed through.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,18 @@ project(SLEEF
 	VERSION ${SLEEF_VERSION}
 	LANGUAGES C)
 
+###############
+### Install ###
+###############
+set(PN sleef)
+set(generated_dir "${CMAKE_CURRENT_BINARY_DIR}/generated")
+set(config_install_dir "lib/cmake/${PN}")
+set(include_install_dir "include")
+set(version_config "${generated_dir}/${PN}ConfigVersion.cmake")
+set(project_config "${generated_dir}/${PN}Config.cmake")
+set(targets_export_name "${PN}Targets")
+set(namespace "${PN}::")
+
 # Sanity check for in-source builds which we do not want to happen
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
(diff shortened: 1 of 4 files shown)


shibatch:
@xsacha Thank you for submitting a patch.
Unfortunately, the patched source code cannot be built.
Please check the output from CI server and fix the problem.

https://travis-ci.org/shibatch/sleef/jobs/501824573

xsacha:
Hey, sorry, I am confused by this option:
option(BUILD_LIBM "libsleef will be built." ON)
I don't think the text is right for this. It seems to always build this library regardless.

shibatch:
Why do you need to install ${TARGET_LIBSLEEF}, ${TARGET_LIBSLEEFGNUABI} and ${TARGET_LIBDFT} at the end of the top CMakeLists.txt?

I thought what you need to additionally install is *.cmake.

xsacha:
Yeah, that's how it works. I give it the TARGET and it brings in all associated libraries, dependencies, includes (unless you did them manually somewhere like the find_package).

That first attempt just didn't work because of a bug/design in that specific CMake yet it worked fine in mine (latest).

This new version should work fine in all CMake.

shibatch:
@xsacha I know I am late, but I am now looking into this PR.
Could you tell me how those atmark-surrounded variables work?

xsacha:
@shibatch when the file is configured, as in `configure_package_config_file`, the @variables@ are replaced by the CMAKE variables with the same name.

shibatch:
Looks good, but I think we need some kind of a tester for making sure that this functionality is not broken.

shibatch:
@fpetrogalli How do you think? If you are happy with this patch, I will merge this and then add a tester.

friendlyanon:
@shibatch #412 superseded this PR

blapie:
I'm closing this PR for the reason pointed by @friendlyanon.

PR #250, opened 2019-03-13
Export defines PUBLIC so that headers work properly

xsacha:
COMMON_TARGET_DEFINITIONS needs to be PUBLIC on the built libraries so that it is exported to the sleefConfig.cmake.

Otherwise, the sleef.h header does not work properly.
SLEEF_STATIC_LIBS and ENABLEFLOAT128 are two examples that need to be exported. I am unsure if all defines need to be exported but currently the defines are not distinguished in any way.

### DIFF
--- a/src/libm/CMakeLists.txt
+++ b/src/libm/CMakeLists.txt
@@ -184,14 +184,14 @@ set_target_properties(${TARGET_LIBSLEEF} PROPERTIES
 )
 
 target_compile_definitions(${TARGET_LIBSLEEF}
-  PRIVATE DORENAME=1 ${COMMON_TARGET_DEFINITIONS}
+  PRIVATE DORENAME=1 PUBLIC ${COMMON_TARGET_DEFINITIONS}
 )
 
 if(COMPILER_SUPPORTS_FLOAT128)
   # TODO: Not supported for LLVM bitcode gen as it has a specific compilation flags
   target_sources(${TARGET_LIBSLEEF} PRIVATE sleefqp.c)
   target_compile_definitions(${TARGET_LIBSLEEF}
-    PRIVATE ENABLEFLOAT128=1 ${COMMON_TARGET_DEFINITIONS})
+    PUBLIC ENABLEFLOAT128=1)
 endif()
 
 if(COMPILER_SUPPORTS_BUILTIN_MATH)
--- a/src/dft/CMakeLists.txt
+++ b/src/dft/CMakeLists.txt
@@ -351,5 +351,9 @@ set_target_properties(${TARGET_LIBDFT} PROPERTIES
 
 # Install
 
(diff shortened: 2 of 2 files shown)


src/libm/CMakeLists.txt
```diff
@@ -184,14 +184,14 @@ set_target_properties(${TARGET_LIBSLEEF} PROPERTIES
...
 target_compile_definitions(${TARGET_LIBSLEEF}
-  PRIVATE DORENAME=1 ${COMMON_TARGET_DEFINITIONS}
+  PRIVATE DORENAME=1 PUBLIC ${COMMON_TARGET_DEFINITIONS}
 )
 
 if(COMPILER_SUPPORTS_FLOAT128)
   # TODO: Not supported for LLVM bitcode gen as it has a specific compilation flags
   target_sources(${TARGET_LIBSLEEF} PRIVATE sleefqp.c)
   target_compile_definitions(${TARGET_LIBSLEEF}
-    PRIVATE ENABLEFLOAT128=1 ${COMMON_TARGET_DEFINITIONS})
```
xsacha: COMMON_TARGET_DEFINITIONS was removed from this line because it is already added in the previous target_compile_definitions

fpetrogalli:
Hi @xsacha , could you please explain what are you trying to fix here? You say that "sleef.h doesn't work properly". What do you mean by that? What is this patch allowing you to do with sleef.h that was not possible to do before?

xsacha:
What I mean is that if you include the header, it will not have the same defines that were used at compile time.

fpetrogalli:
> if you include the header, it will not have the same defines that were used at compile time

Hi , thank you for explaining, but this patch is still not cleat to me. 
Why would sleef.h need to have the same defines that are used at build time? Becasue from what you are saying, I understand that you are thinking of having those defines in sleef.h, did I get it right? If that's the case, could you provide and example when this is needed?

Apologies for asking you to explain more, but I am not a cmake expert and I really don't see why we should export those defines.

xsacha:
Any defines provided through the CMake are not sent when linking the project unless you use the correct CMake functions.

fpetrogalli:
>  By making it PUBLIC, anyone who externally links to the project will also get the defines.

Does this mean that you need to gives visibility to those defines to other CMake projects that include SLEEF as a subproject? Is this what you mean by "externally link"? All these defines are doing is just selecting how to recompile the sources, why would an external project need to see them?

xsacha:
Because when I tried to use sleef headers, it threw undefined linker errors because variables such as avx and float128 were defined at compile time but not when I linked to it in CMake.

This commit fixed these errors.

shibatch:
@fpetrogalli How about this patch? I think this patch is safe to merge.

Issue #249, opened 2019-03-12
Unable to crosscompile without first building natively


xsacha:
I feel there's something wrong with the usage of the build tool when I have to make a complete native build prior to cross compiling.
I found this script which shows the correct way to cross compile this repository: https://github.com/shibatch/sleef/blob/master/travis/before_script.aarch64-gcc.sh

The alternative, which other repositories are able to do, is to use CMake commands where possible to replace executing binaries. Also, you can create host binaries where need be and use them directly.

shibatch:
It is because we migrated from makefile to cmake that the building process is not fully optimized for cmake.
I was also considering replacing generators like mkrename to python scripts, but I haven’t done it because generators are not the main reason we are building natively before starting cross compiling. It’s tester. SLEEF depends on libmpfr for testing.
If we don’t need to test after compilation, then we can substitute generators with scripts.

yuyichao:
It'll also be nice to have some instructions for cross-compilation. I spent a few hours yesterday before getting all the right ingredients in place (a pointer in the README to the travis script and maybe a mention of the `NATIVE_BUILD_DIR` variable would have been enought for me....)

shibatch:
Could you explain why you need cross compilation?
Generally speaking, cross compilation is not easy.
It requires many bug-workarounds.

xsacha:
Need cross compilation so I can compile this for iOS and Android. None of the other repositories have any issues doing this.

shibatch:
iOS and Android would be worth supporting officially.
Let me think.

shibatch:
Android is supported by Travis.
@xsacha Are you interested in also contributing CI setting for Android, and possibly iOS?

https://docs.travis-ci.com/user/languages/android/

xsacha:
I know it's supported by travis and it's supported by this repository but it's just silly the way it forces you to compile a native build prior to cross compiling. That needs to be fixed.

shibatch:
I would say it is not a large problem that cross compilation requires a prior native build.
In order to do testing, this is a necessary process.
Are you thinking about the case where sleef has to be built as a submodule of another project?

xsacha:
Yes, specifically, I am building sleef as a component of pytorch which is built as a component of another project.
Build tools do not usually assume you have to build a package natively prior to cross compiling and then provide the native directory to the CMake in order to build it.

I have over 100 CMake packages in a single project being cross-compiled for iOS, Android, Windows and Linux and this is the first one that has required this.

[...]

PR #251, opened 2019-03-14
Add a CMake generator for target 'alias', removing a binary

xsacha:
This is the first step in a slightly long path of removing compiled binaries used as part of the build process.
This performs the generation of mkalias and most of this code will be reused to do mkrename et al.

When finished, we will be able to cross-compile entirely without native builds, allowing CMake builds to work directly and as part of other packages and build tools.

See: https://github.com/shibatch/sleef/issues/249


fpetrogalli-arm:
Hi @xsacha , thank you for working on this. To my understanding your aim is to simplify the cross compilation process by removing the `mk*` executables used in the build process.

I fear this is not going to be a small piece of work, that will touch a large part of the cmake configuration.

Because of this, I want to propose you to discuss the approach as an issue before proceeding implementing a solution.

If we agree on the design of your solution, it will be easier to review your code, and it reduces the changes of wasting your time.

Please don't think I am trying to push your work back! I fully support the idea of removing the build executables.

Thank you!

xsacha:
Yes, indeed. This was the smallest one to replace and should give you an idea of what I am looking to do (I'm not working on replacing anything else yet).

shibatch:
I think it is easier to replace those generators with python scripts.
Is there any reason that you are trying to do everything with cmake?

xsacha:
The variables exist in CMake and can be used directly rather than discovering your python executable, making sure you find python3 instead of python2, writing the variables somewhere python can see them and running the script.

shibatch:
Please go on.
Make an issue first.

xsacha:
The issue was https://github.com/shibatch/sleef/issues/249

fpetrogalli-arm:
> What @fpetrogalli-arm means is that we should discuss detailed design of your change first.

Yes, this is what I meant, and we can use this patch as a practical example.

 > [...] We need to know how cmakelists will look like after your modifications.

This patch is a good example of what we can expect when converting the `mk*` tools to cmake. Looking at the code, I am not sure this is the best route to take in removing he executable dependencies.

I am particularly concerned about the way we have to specify things like ﻿FUNCLIST. The CMake syntax in this case is not very nice...

@xsacha , would it be possible for you to create also a python version of the patch? I know adding python as a requirement to the build system is requiring more than cmake itself, but I think that:

1. python is available on the majority (if not all) the targets OSes we compile for.
2. python will be easier to maintain/extend than cmake.

PR #252, opened 2019-04-15
Allow sleef to be built as submodule

nicolone:
Sleef doesn't support nesting. The fix is similarly to https://ceres-solver-review.googlesource.com/c/ceres-solver/+/9780:

Using sleef_[SOURCE/BINARY]DIR (which are defined by CMake when
project(sleef) is called, in favour of CMAKE[SOURCE/BINARY]_DIR
enables sleef to be nested within (and built by) a larger CMake
project (which also contains other projects).
CMAKE_[SOURCE/BINARY]_DIR always refers to the top-level source
and binary directories (i.e. the first encountered), as a result if
sleef is a nested project within a larger project, these would not
correctly identify the source/binary directories for sleef (as they
would refer to the root project in which sleef is nested).
Using sleef_[SOURCE/BINARY]_DIR should ensure that sleef always uses
the correct source/binary directories, irrespective of whether sleef
is nested or not.

See https://github.com/pytorch/pytorch/issues/17181


CMakeLists.txt
```diff
@@ -32,8 +32,15 @@ project(SLEEF
 	VERSION ${SLEEF_VERSION}
 	LANGUAGES C)
 
+if(NOT DEFINED sleef_SOURCE_DIR)
```
fpetrogalli-arm: Two requests:

1. Please add a comment that refers to the cmake documentation that describes the auto-generated variable name.
2. Is there a way to build this variable name programmatically? What happens if we decide to change the name of the project, or a user decides to change the name of the top level folder of sleef?

xsacha: Doesn't CMAKE already have support for this with CMAKE_CURRENT_SOURCE_DIR or even CMAKE_CURRENT_LIST_DIR

It also has CMAKE_CURRENT_BINARY_DIR

PR #266, opened 2019-07-26
Fix compilation under Windows with MinGW

nxrighthere:
The compiler-specific condition for `type` and `cat` is incorrect. Also `<io.h>` is missed for `_get_osfhandle`.

### DIFF
--- a/src/common/arraymap.c
+++ b/src/common/arraymap.c
@@ -28,6 +28,7 @@ static FILE *OPENTMPFILE() { return tmpfile(); }
 static void CLOSETMPFILE(FILE *fp) { fclose(fp); }
 #else
 #include <Windows.h>
+#include <io.h>
 
 static void FLOCK(FILE *fp) { }
 static void FUNLOCK(FILE *fp) { }
--- a/src/libm/CMakeLists.txt
+++ b/src/libm/CMakeLists.txt
@@ -22,7 +22,7 @@ foreach(SIMD ${SLEEF_HEADER_LIST})
   list(APPEND SLEEF_HEADER_COMMANDS COMMAND $<TARGET_FILE:${TARGET_MKRENAME}> ${HEADER_PARAMS_${SIMD}} >> ${SLEEF_INCLUDE_HEADER})
 endforeach()
 
-if(MSVC)
+if(WIN32)
   string(REPLACE "/" "\\" sleef_footer_input_file "${SLEEF_ORG_FOOTER}")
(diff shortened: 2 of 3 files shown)


nxrighthere:
What's wrong with the Jenkins? The first commit with `WIN32` didn't pass the build, the second is pending for almost 5 hours.

shibatch:
@nxrighthere Thank you for your commit!
I am sorry, but Jenkins is not working properly now, due to expiration of emulator.
I will fix it soon.

yuyichao:
I'm not sure what's the correct condition but this breaks my compilation under msys2 using `makepkg-mingw` for `mingw64` ....

nxrighthere:
On which OS?

yuyichao:
Windows 10

yuyichao:
[This](https://github.com/nacs-lab/nacs-pkgs/blob/c84ada62e2d64efd0ce34f4ce4c4dda1b50a146f/windows/mingw-w64-sleef-git/PKGBUILD) was the `PKGBUILD` (the master branch has one additional patch to revert this change).

The `build.make` has a line like `type C:\X\Y\Z >> ...` and the `type` gives an error about not being able to fine `C:XYZ`.

nxrighthere:
I think the condition should be then `if(MSVC OR MINGW AND WIN32 AND NOT MSYS)` to separate standalone MinGW and the one under MSYS2.

yuyichao:
That should cover my case for sure.

However, I don't really use windows and have no idea what's the difference between all these posix like environment for windows so I can't say what should be the correct logic ...... = = ....

nxrighthere:
If you use the standalone MinGW then `cat` is not available since it's a UNIX utility and `type` works just fine. In your case I think `cat` is available under MSYS2 so we just need to separate the two.

yuyichao:
Sounds fair. What I mainly don't understand is why does `type` **not** work for me if it works with the standalone mingw.

nxrighthere:
Good question, I'm not using MSYS2, so can't explain where the problem is.

yuyichao:
In particular, for me, the `build.make` was executed by `/bin/sh`, which use the shell builtin `type` command rather than the windows cmd command (the error message for file not found are different and the shell builtin one also doesn't support `C:\...` paths somehow...). The command line argument given to it is also escaped incorrectly causing all the `\` to be missing from the file name....

I just tried using the `mingw-w64-x86_64-make` package instead of the `make` package for the `make` (`mingw32-make`) command and it didn't make a difference..... (they both pick up `/bin/sh`)

shibatch:
One solution could be to build a command for displaying the content of a file from a source code.
I tried to do that before, but this proposal was rejected.

https://github.com/shibatch/sleef/pull/138

PR #274, opened 2019-11-22
Build on OpenBSD

xg6:
Dear Prof. Shibata, 

First of all thank you for providing this library. I am using it with PyTorch on OpenBSD.

With this simple fix your library builds without errors on OpenBSD and all tests pass except for the ilogb denormal/nonnumber tests where "correct" return INT_MIN for inf arg. This appears to be due to a bug in the OpenBSD amd64 implementation of ilogb as the [documentation](https://man.openbsd.org/ilogb) states that the function should return INT_MAX for inf arg. Tests return INT_MAX.

Best regards,
XG6


shibatch:
Hello,

Thank you for your commit and I am sorry for delayed response.
The patch looks good to me.

PR #275, opened 2019-12-04
MSVC static library changes

josephalvis:
This patch has two small changes to improve building and linking to sleef as a static library on windows:

1. Add SLEEF_STATIC_LIBS to sleef-config.h so that it is automatically set without using a define before including sleef.h
2. Enable setting the runtime library via CMAKE_MSVC_RUNTIME_LIBRARY. This is necessary when using anything other than the default runtime library to avoid undefined symbols. e.g. DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded


shibatch:
We do not have very good knowledge on MSVC.
Please explain why these two items are necessary.
Do you mean that you are having problems when compiling with -MT option?

josephalvis:
Hi,
Yes these are the issues I faced:

1 - When linking to a static build of sleef, all of the symbols were undefined. It turns out that SLEEF_STATIC_LIBS changes the definition of IMPORT/EXPORT (i.e. declspec(dllimport)). That means the client code needs to define/undefine SLEEF_STATIC_LIBS depending on the way sleef was built. Adding SLEEF_STATIC_LIBS to sleef-config.h eliminates the need to do that on the client side.

2 - When linking to a static build of Sleef, the selected runtime libraries need to match  - otherwise the linker complains about missing functions like imp_malloc. That is because by default, Sleef will link to MultiThreadedDll (/MD using _declspec(dllimport)). If the client library is linking to a static runtime library, the symbol name is actually different (i.e. malloc instead of imp_malloc).

If the client code is linking to MultiThreadedDll (the default), there is no issue and nothing needs to be set. 

The CMAKE_MSVC_RUNTIME_LIBRARY setting simply allows the user to change the Sleef linkage when something other than the default is needed.

TLDR: I want to build Sleef staticly. I am also linking to a static runtime library (/MT) while sleef links to a shared runtime library (/MD) by default. The default behaviour is unchanged but now /MT can be selected by defining CMAKE_MSVC_RUNTIME_LIBRARY on the command line.

For reference:
https://cmake.org/cmake/help/git-stage/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html

shibatch:
Rebased https://github.com/shibatch/sleef/pull/330

PR #278, opened 2019-12-30
Fixes FLAGS_ENABLE_SSE<n> flags for MSVC 32-bit build

WSSDude:
In MSVC, /arch:SSE2 is default only for 64-bit builds. It should therefore be set for both SSE2 and SSE4 in 32-bit build.


shibatch:
@WSSDude Thank you for making the pull requests.
All PRs are now merged.

PR #279, opened 2019-12-30
Removes disabling of OpenMP on MSVC

WSSDude420:
OpenMP is supported (2.0 specification) in MSVC from [2015](https://docs.microsoft.com/en-us/cpp/parallel/openmp/openmp-in-visual-cpp?view=vs-2015).

This therefore removes disabling of OpenMP on MSVC. Code compiles without issues, tested on VS2017, VS2019 in both, 32-bit and 64-bit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,8 +155,3 @@ if(SLEEF_SHOW_CONFIG)
     message(STATUS "Building SLEEF with AArch64 Vector PCS support")
   endif()
 endif(SLEEF_SHOW_CONFIG)
-
-if (MSVC)
-  message("")
-  message("*** Note: Parallel build is not supported on Microsoft Visual Studio")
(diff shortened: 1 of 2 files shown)


WSSDude:
Hm, this seems weird to me really...
`LINK : warning LNK4044: unrecognized option '/openmp'; ignored`

I tested it on all compilers I said. Isn't it problem then with some specific versions? Because it should be supported for quite some time now and it seems to work without issues (at least for me).

First time seeing this warning pop-up and seeing tests actually fail.

From the version number, this seems to be VS2017 you are using on your test server from what I can see. We started utilizing SLEEF somewhere like 2 months ago at our work on VS2017 too with enabled OpenMP and this did not happen. Niether the warning or failed tests. 

Maybe outdated compiler version is in place on the server?

Issue #282, opened 2020-02-18
Add generators of headers for inlining whole sleef functions


shibatch:
There have been several requests for providing vectorized SLEEF functions in a header file,

https://github.com/shibatch/sleef/issues/230

It turned out that this functionality is not too hard to implement.
We can use C preprocessor (cpp) to generate such header files.
We will apply cpp a few times to sleefsimddp.c and sleefsimdsp.c with SLEEF_GENHEADER macro and other appropriate macros defined.
Part of the codes in misc.h and helper files will be guarded with #if, so that irrelevant lines will not appear in the resulting header files.

A header file is generated for each vector extension.
The resulting header files contain inline functions defined in dd.h and df.h, and functions defined in sleefsimddp.c and sleefsimdsp.c with all the function names renamed to the ones specified in the standard SLEEF API.

Since the inline functions in helper files are not renamed, those header files with different vector extensions cannot be included from a same file.

An example of generated header file can be seen at the following URL.

http://www.aist-nara.ac.jp/~n-sibata/header_example/sleefinline_avx2.h

Please see the example of generated header file and let me know you thought.

shibatch:
Below is another example.

http://www.aist-nara.ac.jp/~n-sibata/header_example/sleefinline_sse2.h
http://www.aist-nara.ac.jp/~n-sibata/header_example/helloinline.c
http://www.aist-nara.ac.jp/~n-sibata/header_example/helloinline.s

Please see that the function is inlined in helloinline.s.

fpetrogalli:
This sounds good to me in principle. One thing that you should make sure, it that duplication is minimal, if any is needed at all. This is to avoid having to update code in both the library and the header files when an algorithm is changed. Does that make sense?

shibatch:
There will be almost no code duplication. Those header files are generated from the existing sleefsimddp.c and sleefsimdsp.c. Some of the macros have to be written in a special format in order to process them in a later stage.

fpetrogalli:
Sounds good to me!

blapie:
I'm closing this issue as I believe it was solved by #283. Feel free to re-open if not the case or open a new issue.

Issue #230, opened 2018-11-05
Sleef functions can't be inlined


colesbury:
In PyTorch, we'd like to use Sleef's vectorized implementation of elementary functions as building blocks. For example, we'd like to implement a vectorized [`sigmoid()`](https://en.wikipedia.org/wiki/Sigmoid_function) function using `exp`. However, calling into Sleef's exp() is expensive because it incurs the cost of a non-inlineable function call.

It would be great if Sleef provided the instruction-set specific vectorized functions in a header file, or some other way that can be inlined by the compiler.

fpetrogalli-arm:
Hi @colesbury ,
what compiler do you use to build PyTorch? I am asking because SLEEF can be built as an LLVM bytecode library. If you rely on clang in your tool-chain, it should be possible to inline all SLEEF functions.

colesbury:
We support GCC, clang, and MSVC. We can't rely on LTO.

fpetrogalli-arm:
> We support GCC, clang, and MSVC. We can't rely on LTO.

The clang mechanism wouldn't rely on LTO. In any case, LLVM bitcode is not viable because you have to support also GCC and MSVC.

Would adding `sigmoid()` to SLEEF itself help your specific case? Or do you have many functions you would like to add?

shibatch:
Adding VML-style API with all math functions inlined is not impossible, but it would take some time to implement.

shibatch:
@colesbury I wrote a prototype of SleefVML. Please check my gist and let me know if this is what you want.

https://gist.github.com/shibatch/70f48e153ecc9d586386dc1676d564cc

In this file, a few activation functions derived from the Wikipedia entry are implemented.
https://en.wikipedia.org/wiki/Activation_function

These functions can be computed with a VML-style API.
```
void SleefVML_vmsSigmoid(size_t n, float *arg0, float *ret, uint64_t mode);
```

With optimization, all the functions will be vectorized.

```
g++ -fopenmp -Wno-attributes -I./include -L./lib -march=avx2 -O3 sleefvml_test.cpp
```

For example, the vectorized assembly output of the sigmoid function is as follows.
```
.L50:
        vmovups 0(%r13,%r15), %xmm0
        vinsertf128     $0x1, 16(%r13,%r15), %ymm0, %ymm0
        vmovaps %ymm1, -80(%rbp)
        vxorpd  .LC3(%rip), %ymm0, %ymm0
        call    Sleef_expf8_u10avx2
[...]

PR #283, opened 2020-02-21
Generate inline headers

shibatch:
This patch implements the functionality described in issue https://github.com/shibatch/sleef/issues/282.

The header files for inlining sleef functions are generated if -DBUILD_INLINE_HEADERS=TRUE is specified as a cmake option.

In order to use one of these headers, the following 3 macros have to be defined.

* SLEEF_ALWAYS_INLINE : This is the function attribute specified for helper functions
* SLEEF_INLINE : This is the function attribute specified for SLEEF functions.
* SLEEF_CONST : This is usually const attribute.

Each header file corresponds to one vector extension. Only one of the header files can be included from the same source file.

It also builds libsleefinline.a, that is referred from the inlined functions. 

Aarch32 is not supported at this time.

--

The main change to the source codes can be summarized into the following two points.
1. src/libm/CMakeLists.txt : This patch adds instructions for building the header files. It applies C preprocessor and sed command multiple times to generate the header files.
2. Addition to helper files : This patch adds lines beginning with //@ which is specially treated during generation of the header files.

It seems that parallel build does not work even on linux computers. According to the following page, parallel build is always unsafe if multiple COMMANDs are used in add_custom_command. Accordingly, I changed the CI settings to remove parallel builds.

https://stackoverflow.com/questions/60105230/how-to-run-cmake-commands-in-add-custom-command-in-order

Currently, Jenkins servers are not available. I hope they will be available again before this patch is approved. I manually ran the builds and tests on every platform.

@colesbury Github does not allow me to add you as a reviewer, but please tell us your thoughts.


colesbury:
Thanks, I will test this out today

colesbury:
Thanks, this addresses the request in https://github.com/shibatch/sleef/issues/230 

In my limited testing, the functions are now successfully inlined at call sites. In our sigmoid function, I measured a ~14% improvement for 1 million floats by inlining the call to Sleef_expf8_u10. I haven't done measurements of other functions.

Currently, we include Sleef as a submodule and build it as part of the PyTorch build process. The changes to the build make this more difficult. I think we may switch to building Sleef separately and just committing the artifacts (e.g. sleefinline_avx2.h) to the PyTorch repo. I think that should be fine.

shibatch:
Good.
Please credit SLEEF if the generated source code will be mixed into the source tree of PyTorch.

src/libm-tester/iutsimd.c
```diff
@@ -176,68 +248,74 @@ typedef Sleef___m256_2 vfloat2;
...
 typedef Sleef_float_2 vfloat2;
 #endif
+#endif
 
 //
 
 #ifdef ENABLE_DP
-int check_featureDP() {
-  if (vavailability_i(1) == 0) return 0;
+int check_featureDP(double d) {
```
fpetrogalli: Is this needed for the header functionality? Same question for the float version below.

src/libm/CMakeLists.txt
```diff
@@ -288,6 +290,51 @@ foreach(SIMD ${SLEEF_SUPPORTED_EXTENSIONS})
...
+    foreach(SIMD ${SLEEF_SUPPORTED_EXTENSIONS})
+      if(COMPILER_SUPPORTS_${SIMD})
+	string(TOLOWER ${SIMD} SIMDLC)
+
+	set(INLINE_HEADER_FILE ${PROJECT_BINARY_DIR}/include/sleefinline_${SIMDLC}.h)
+	add_custom_command(
+	  OUTPUT ${INLINE_HEADER_FILE} 
+
+	  COMMAND echo Generating sleefinline_${SIMDLC}.h
+	  COMMAND "${CMAKE_C_COMPILER}" ${FLAG_PREPROCESS} ${FLAG_PRESERVE_COMMENTS} ${FLAG_INCLUDE}${PROJECT_SOURCE_DIR}/src/common ${FLAG_INCLUDE}${PROJECT_SOURCE_DIR}/src/arch ${FLAG_INCLUDE}${CMAKE_CURRENT_BINARY_DIR}/include/ ${FLAG_DEFINE}SLEEF_GENHEADER ${FLAG_DEFINE}ENABLE_${SIMD} ${FLAG_DEFINE}DORENAME ${CMAKE_CURRENT_SOURCE_DIR}/sleefsimddp.c > ${CMAKE_CURRENT_BINARY_DIR}/sleef${SIMD}.h.tmp1
```
fpetrogalli: Please break long lines. It is hard to understand what is going on here. Please try not to pass the 80 chars limit.

src/libm/rempitab.c
```diff
@@ -5,7 +5,13 @@
 
 #include "misc.h"
 
-NOEXPORT ALIGNED(64) const double rempitabdp[] = {
+#if !defined(SLEEF_GENHEADER)
+#define FUNCATR NOEXPORT ALIGNED(64)
+#else
+#define FUNCATR EXPORT ALIGNED(64)
+#endif
```
fpetrogalli: Why is this needed?

src/libm/sleefsimdsp.c
```diff
@@ -22,7 +24,11 @@ extern const float rempitabsp[];
 
 #ifdef ENABLE_SSE2
 #define CONFIG 2
+#if !defined(SLEEF_GENHEADER)
 #include "helpersse2.h"
+#else
+#include "macroonlySSE2.h"
+#endif
```
fpetrogalli: Do we need to header files? Couldn't we  conditionally compile the same header file, exposing only the macros when SLEEF_GENHEADER is defined? This could avoid the need of generating the `macroonly*` files.

travis/before_script.osx-clang.sh
```diff
@@ -2,4 +2,4 @@
 set -ev
 mkdir sleef.build
 cd sleef.build
-cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../install -DSLEEF_SHOW_CONFIG=1 -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl -DENFORCE_TESTER3=TRUE -DBUILD_QUAD=TRUE ..
+cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../install -DSLEEF_SHOW_CONFIG=1 -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl -DENFORCE_TESTER3=TRUE -DBUILD_QUAD=TRUE -DBUILD_INLINE_HEADERS=TRUE ..
```
fpetrogalli: Do I understand correctly that you are always generating the header files because you want to test them on each architecture with the `iuti*` executables? Has this increased the running time in CI?

fpetrogalli:
How comes that CI hasn't completed yet?

shibatch:
Because of a policy change in network security at my institute, I now have to move the CI servers to a different network segment.
I had a meeting with the people in charge of network management, and I am told to change the network settings of the servers.
However, they did not know what CI servers are, and I was asked to have a meeting with them again.
In the second meeting, I told them how the servers work, and they are now saying that additional security measures are required.

Jenkinsfile
```diff
@@ -15,10 +15,10 @@ pipeline {
  			 mkdir build
 			 cd build
 			 cmake -DCMAKE_INSTALL_PREFIX=../install -DSLEEF_SHOW_CONFIG=1 -DENFORCE_TESTER3=TRUE -DBUILD_QUAD=TRUE ..
-			 make -j 6 all
+			 make -j 1 all
```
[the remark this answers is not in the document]
shibatch: Because add_custom_command does not support parallel builds. In this patch, many COMMANDS are executed within one add_custom_command, and parallel builds fail even on linux computers.

https://stackoverflow.com/questions/60105230/how-to-run-cmake-commands-in-add-custom-command-in-order

src/arch/helperadvsimd.h
```diff
@@ -9,27 +9,38 @@
...
 
+#if !defined(SLEEF_GENHEADER)
 #include <arm_neon.h>
 #include <stdint.h>
 
 #include "misc.h"
+#endif // #if !defined(SLEEF_GENHEADER)
 
 #define ENABLE_DP
+//@#define ENABLE_DP
```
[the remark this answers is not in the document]
shibatch: Please read my first comment. Generation of the header files requires multiple processing with cpp, and some of the macros are required in later use.

Addition to helper files : This patch adds lines beginning with //@ which is specially treated during generation of the header files.

src/libm-tester/iutsimd.c
shibatch: With the headers, functions like this have greater tendency of being optimized away.
This change is needed to prevent optimizer from removing this function.

src/libm/rempitab.c
shibatch: This table is referenced from the source code including the generated header.
So, this table has to be compiled into a library.

src/libm/sleefsimdsp.c
shibatch: Instead of only exposing some macros, I chose the approach to define some of the macros beginning with //@.

Below is comparison between pros and cons.
Approach 1: Exposing a part of the macros according to whether SLEEF_GENHEADER is defined.
Pros: The mechanism is straightforward.
Cons: We need to protect most of the part in the macro with large #ifdef. This is hard to read.
Generation of macros is not very understandable anyway.

Approach 2: Define some of the macros beginning with //@
Pros: Easier to read and manage the source code.
Cons: This is a dedicated way for SLEEF. and requires understanding how the macros are generated.

travis/before_script.osx-clang.sh
shibatch: Yes, running time in CI is increased.
If that is a problem, we can just turn off generation.

src/arch/helperavx.h
```diff
@@ -649,6 +664,7 @@ static INLINE vargquad vcast_aq_vm2(vmask2 vm2) {
   return a;
 #endif
 }
+#endif
```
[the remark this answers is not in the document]
shibatch: Done.

src/libm/CMakeLists.txt
shibatch: I added comments in this section.

PR #288, opened 2020-03-01
PPC64LE: Fix build issue on gcc

seiko2plus:
#### Several improvements for VSX
  - fix the build on gcc
  - fix accuracy tests
  - fix the build with C++ when 'vector' compiler token is combined,
    NOTE: the generated header is using '__vector' instead of 'vector'
  - fix unaligned memory load
  - set rounding to the nearest even, make it equivalent to X86
  - improve the performance of conversions
  - improve the performance of shuffles and reverses
  - cleanup the whole VSX implementation

#### TODO
- [x]  activate Travis on ppc64le
- [x] fix overflow
- [x] fix  test failures on Clang
- [x] finalize and testing Sleef on a wide range of compilers versions

### DIFF
--- a/src/arch/helperpower_128.h
+++ b/src/arch/helperpower_128.h
@@ -6,7 +6,7 @@
 #if CONFIG == 1 || CONFIG == 2
 
 #ifndef __VSX__
-#error Please specify -mvsx.
+#error Please specify -mcpu=power8 or -mcpu=power9
 #endif
 
 #else
@@ -15,11 +15,11 @@
 
 #define ENABLE_DP
 #define LOG2VECTLENDP 1
-#define VECTLENDP (1 << LOG2VECTLENDP)
+#define VECTLENDP 2
 
 #define ENABLE_SP
-#define LOG2VECTLENSP (LOG2VECTLENDP+1)
-#define VECTLENSP (1 << LOG2VECTLENSP)
+#define LOG2VECTLENSP 2
+#define VECTLENSP 4
 
 #if CONFIG == 1
 #define ENABLE_FMA_DP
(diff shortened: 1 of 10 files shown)


edelsohn:
This is a great improvement. I believe that the src/libm-tester/tester3.c patch also is necessary.

edelsohn:
More specifically, I agree that the current test for VSX support is wrong.

I'm not certain that Sleef should impose -mcpu=power8 because that inhibits improvements in future processors.

The problem with libm-tester/tester3.c is:

1) It should use `__vector`
2) The vector should be constructed with { } not ( ), e.g.

` return (__vector double) { a[0], a[1] }; }`
and
`return (__vector float) { a[0], a[1], a[2], a[3]; }`

shibatch:
Could you also add travis configuration?

seiko2plus:
@edelsohn, I fixed the 'tester3.c'. it is safe to use flag `-mcpu=power8`, since it will only run with ppc64le, also it provides a binary safety in case of running it on both Power8 & Power9.  Later Sleef should add support for VSX3 with compiler flag `-mcpu=power9` to activate it.
 
@shibatch, Yes sure I will

src/arch/helperpower_128.h
```diff
@@ -53,29 +72,42 @@ static INLINE int vavailability_i(int name) { return 3; }
...
-static void vstoreu_v_p_vi2(int32_t *p, vint2 v) { vec_st(v, 0, p); }
-static vint vloadu_vi_p(int32_t *p) { return vec_ld(0, p); }
-static void vstoreu_v_p_vi(int32_t *p, vint v) { vec_st(v, 0, p); }
-
-static INLINE vdouble vload_vd_p(const double *ptr) { return (vector double)vec_ld(0, (const int *)ptr); }
-static INLINE void vstore_v_p_vd(double *ptr, vdouble v) { vec_st((vector int)v, 0, (int *)ptr); }
-static INLINE vdouble vloadu_vd_p(const double *ptr) { return (vector double) ( ptr[0], ptr[1] ); }
-static INLINE void vstoreu_v_p_vd(double *ptr, vdouble v) { ptr[0] = v[0]; ptr[1] = v[1]; }
+static vint2 vloadu_vi2_p(int32_t *p)
+{ return vec_vsx_ld(0, p); }
```
seiko2plus: @edelsohn, @shibatch, Casting can be used for loading from memory, e.g. `*((vec_double2*)p)`, it doesn't work on CLANG and it seems works on GCC.

edelsohn: Vector assignment should work with Clang.  Can you provide a specific example of an idiom that doesn't work? And by doesn't work do you mean syntax error, compiler error or wrong code?  If one lies about alignment with type punning it can produce incorrect code with any compiler.

edelsohn:
Is it not possible to use assignment and remove all of the references to vec_vsx_ld and vec_vsx_st?  It really is better to not use those intrinsics.  Maybe it is better to require a minimum version of GCC instead of trying to support ancient versions for backward compatibility. The backward compatibility may be much worse for performance.

src/arch/helperpower_128.h
seiko2plus: sorry for the delayed response, Load/Store via (cast & dereference) is buggy with almost all clang versions(except clang-4) and gcc8(older versions are working just fine) when VSX3(power9/ISA3.00) isn't enabled, these compilers tend to generate instructions 'lvx/stvx' instead of 'lxvd2x/lxvw4x/stxvd2x/stxvw4x', which lead to break the accuracy tests when unaligned memory addresses are involved.
please check the full report at https://github.com/seiko2plus/vsx_mem_test.

However, I update it the patch to use load/store via (cast & dereference) only when VSX3 is enabled or with GCC versions up to 7 when VSX(2.07) is enabled

seiko2plus:
@edelsohn,
> Is it not possible to use assignment and remove all of the references to vec_vsx_ld and vec_vsx_st?

Only when VSX3 is enabled or up to GCC 7, please check my previous comment.

> Maybe it is better to require a minimum version of GCC instead of trying to support ancient versions for backward compatibility.  The backward compatibility may be much worse for performance.

There's no need it for it, the current workarounds use inline asm and I don't think that will cause any performance lose. however, if you think that may lead to cause performance issue, I can use unexposed built-in intrinsics instead.

shibatch:
Actually, people care pretty much about backward compatibility.
In a case like this, we usually check the compiler support for the corresponding feature in cmake, and use preprocessor macros to enable usage of that feature.
It is okay that you go ahead with replacing those intrinsics with better ones. I will do the compatibility things afterward.

seiko2plus:
@shibatch, the current backward compatibility doesn't affect the performance, I checked all the generated instructions on a wide range of compilers versions, I think this patch is ready to merge.
I think the upcoming patches should focus on :
- supporting VSX3(ISA3.00)
- add support for runtime CPU dispatching
- support gnuabi
- introduce newly mapped intrinsics and provide more compatibility with other SIMD extensions

src/arch/helperpower_128.h
```diff
@@ -31,334 +31,748 @@
...
+//// load
+#if defined(__POWER9_VECTOR__) || (!defined(__clang__) && defined(__GNUC__) && __GNUC__ < 8)
+static vint vloadu_vi_p(const int32_t *ptr)
+{ return *((vint*)ptr); }
+static INLINE vint2 vloadu_vi2_p(const int32_t *ptr)
+{ return *((vint2*)ptr); }
+static INLINE vfloat vloadu_vf_p(const float *ptr)
+{ return *((vfloat*)ptr); }
+static INLINE vdouble vloadu_vd_p(const double *ptr)
+{ return *((vdouble*)ptr); }
```
seiko2plus: @edelsohn, ping

shibatch:
I am interested in supporting gnuabi on POWER platforms.
Do you have any reference for this?

@fpetrogalli Do you have anything to say?

fpetrogalli:
> I am interested in supporting gnuabi on POWER platforms.
> Do you have any reference for this?
> 
> @fpetrogalli Do you have anything to say?

You mean whether there is a Vector Function ABI specification for power that defines a mangling function similar to the ones at [1] and [2]? Sorry I have no idea whether such document exists. 

[1] https://github.com/ARM-software/abi-aa/blob/master/vfabia64/vfabia64.rst#45vector-function-name-mangling (AArch64)
[2] https://software.intel.com/en-us/articles/vector-simd-function-abi (x86)

shibatch:
@edelsohn Is it okay to merge this patch?

edelsohn:
The patch from Sayed is good.

We had a long discussion internally and the suggestion to use pointer assignment was wrong.

The problems highlighted by Sayed's testing aren't bugs in the compiler, but undefined behavior.  One cannot cast from a scalar to a vector because the scalar is not guaranteed to have the necessary alignment.  This was clarified in the ELFv2 ABI Errata

https://openpowerfoundation.org/?resource_lib=openpower-elfv2-errata-elfv2-abi-version-1-4

The current recommendation is indeed to use vec_xl and vec_xst for vector accesses, not vec_ld/st and not vec_vsx_ld/st.  vec_xl/xst will generate better code for unaligned accesses.

seiko2plus:
@edelsohn,
> We had a long discussion internally and the suggestion to use pointer assignment was wrong.
The problems highlighted by Sayed's testing aren't bugs in the compiler, but undefined behavior

my suggestion is (cast and dereference) which isn't undefined behavior, almost all GCC compilers haven't required the pointer to be aligned to 16-byte and it seems the newest versions of GCC don't accept this behavior anymore but only in case if VSX3//POWER9 isn't enabled.

Also if your suggestion of "vector assignment" is to use curly braces initlizer {} e.g. ` __vector float load = (__vector float){ptr[0], ptr[1], ptr[3], ptr[4]}`, well I tested this method in [vsx_mem_test](https://github.com/seiko2plus/vsx_mem_test) and it shows a massive lose of performance in gcc.

> The current recommendation is indeed to use vec_xl and vec_xst for vector accesses, not vec_ld/st and not vec_vsx_ld/st. vec_xl/xst will generate better code for unaligned accesses.

vec_ld/st is only used with aligned memory access and  vec_xl/vec_xst is the new synonym for vec_vsx_ld/st, equivalent to XLC compiler, take a look at the altivec header
https://github.com/gcc-mirror/gcc/blob/884d914111228eed977d794f38e4cc88bf132a58/gcc/config/rs6000/altivec.h#L358-L363

@shibatch,
> I am interested in supporting gnuabi on POWER platforms.
Do you have any references for this?

My bad, I saw about a year ago patches were sent to libc alpha to add VSX support, but now when I checked libc's master branch, I couldn't find any support for it, I will check their maillist archive helpfully to know the reason for not accepting these patches.

edelsohn:
IBM officially has announced that the IBM Compilers are transitioning to an LLVM-based infrastructure:

https://developer.ibm.com/linuxonpower/2020/02/23/ibm-c-c-and-fortran-compilers-to-adopt-llvm-open-source-infrastructure/

I would not worry about the previous XLC Compiler, especially not on Linux.

Technically, vec_xl and vec_xst are defined in the ABI, but reality takes precedence.

I don't understand the questions about gnuabi.

shibatch:
> I don't understand the questions about gnuabi.

In order for a compiler to adopt a vector ABI, that ABI has to be defined. At least the mangling rule for vector functions has to be determined. More specifically, the character that designates the vector extension in POWER architecture has to be determined, at least. This decision is not SLEEF-specific, but comapny-wide.

I would like to know if OpenPOWER or IBM has already made such a decision.

shibatch:
Please make an issue before proceeding with vector-abi support.

edelsohn:
IBM Power Toolchain team has been coordinating with the GCC/GLIBC community on the vector ISA for libmvec.

shibatch:
Okay, the vector ABI is already defined.

seiko2plus:
@edelsohn, compilers treat Altivec `__vector` differently when it comes to pointers, I'm not talking here about 'static cast', In GCC there's no need to add `__attribute__((aligned(1)))` to the pointer, while Clang recommended it but still not necessary.

For example, look at how clang defining `vec_xl` and `vec_xst`:
https://github.com/llvm-mirror/clang/blob/65acf43270ea2894dffa0d0b292b92402f80c8cb/lib/Headers/altivec.h#L16347-L16369

By the way, I tried to use `__attribute__((aligned(1)))` but still accuracy tests keep failing, also `vec_xl` and `vec_xst` in clang failing too due to generating generate instructions 'lvx/stvx' instead of 'lxvd2x/lxvw4x/stxvd2x/stxvw4x'.

Why do I consider it a GCC(>=8)/Clang(>4) bug?
- Because it works in all GCC versions up to 7
- Clang implements `vec_xl` and `vec_xst` through `(cast and dereference)` and still fails,
  while the Clang intrinsics `vec_vsx_ld`, `vec_vsx_st` don't
- It works without any issues when VSX3/ISA3.00 been enabled through `-mcpu=power9` or `-mpower9-vector`

What's the current situation for unaligned memory access when VSX3 isn't enabled?
- Avoid the use of `vec_xl/vec_xst` and pointer casting in clang, and use `vec_vsx_ld/vec_vsx_st` instead
- Avoid the use of pointer casting for the newest versions starting from GCC8, and use `vec_vsx_ld/vec_vsx_st` or `vec_xl/vec_xst` instead.

seiko2plus:
@edelsohn,
> I would not worry about the previous XLC Compiler, especially not on Linux.

As long I know all XLC versions I deal it with are using clang as a backend, not just LLVM,
the current merged patch will not work with XLC, I have to modify several intrinsics related to conversions
and using `vec_xl/vec_xst` for unaligned memory access.

seiko2plus:
@edelsohn, 
> I don't understand the questions about gnuabi.

Sleef provides ELF shared object (.o) that can override libmvec symbols through the linker, in order to use Sleef vector math functions instead of libmvec within OpenMP.

shibatch:
Arm HPC compiler is linking SLEEF with the vector ABI. Actually, this is the main reason that SLEEF supports the vector ABI at this moment.
Some people were working on LLVM side for enabling vector math calls via the vector ABI, but it looks like stalled.
We also need OpenMP support by the compilers in order to truly realize interchangeability between different vector math libs.

edelsohn:
[Power Vector Function ABI](https://github.com/power8-abi-doc/vector-function-abi/)

edelsohn:
@seiko2plus Sayed, Can you open Bugzilla PRs for GCC and Clang to report these issues?  If the documented, official interfaces don't work, they need to be fixed.  We need the issues reported to the compiler communities, not described here and your performance repo.  Thanks!

shibatch:
> Power Vector Function ABI

Thank you!
No vector ABI yet for s390x?

shibatch:
@edelsohn There are two documents describing vector ABI for x86.

* https://software.intel.com/en-us/articles/vector-simd-function-abi
* https://sourceware.org/glibc/wiki/libmvec?action=AttachFile&do=view&target=VectorABI.txt

There is a difference in the characters representing the ISA between these two documents. The first one says its x, y and z. The second one says its b, c, d and e. The real gcc conforms to the second document.

https://godbolt.org/z/nkKVys

So, I think the same character "b" is used both in x86_64 and powerpc.
Is this intentional?

edelsohn:
I don't know about x86-64. I think that libmvec is trying to be consistent.  The Power Vector ABI also is on sourceware wiki.

https://sourceware.org/glibc/wiki/powerarchvectorfuncabi?action=AttachFile&do=view&target=vector-function-abi-final.html

edelsohn:
There is no vector ABI for s390x. I expect that it will follow Power, especially because x86-64 seems to be consistent as well.

seiko2plus:
@edelsohn,
> We need the issues reported to the compiler communities, not described here and your performance repo.

sure, I will inform them

PR #291, opened 2020-03-23
Add IBM System/390 support

shibatch:
This patch adds IBM System/390 support.
Clang is not supported at this time because it seems not supporting VX intrinsics properly.

Note that this is the first big-endian architecture that SLEEF supports.


Configure.cmake
```diff
@@ -54,6 +54,7 @@ set(SLEEF_SUPPORTED_EXTENSIONS
   ADVSIMD ADVSIMDNOFMA SVE SVENOFMA                     # Aarch64
   NEON32 NEON32VFPV4                                    # Aarch32
   VSX VSXNOFMA                                          # PPC64
+  Z13 Z13NOFMA Z14 Z14NOFMA		                # IBM Z
```
seiko2plus: Would it be better if `ZVECTOR` and `ZVECTOR2` are used instead of CPU names?

Configure.cmake
```diff
@@ -600,6 +636,46 @@ if (ENFORCE_VSX AND NOT COMPILER_SUPPORTS_VSX)
...
+# IBM Z
+
+option(DISABLE_Z13 "Disable Z13" OFF)
+option(ENFORCE_Z13 "Build fails if Z13 is not supported by the compiler" OFF)
+
+if(SLEEF_ARCH_S390X AND NOT DISABLE_Z13)
+  set (CMAKE_REQUIRED_FLAGS ${FLAGS_ENABLE_Z13})
+  CHECK_C_SOURCE_COMPILES("
+  #include <vecintrin.h>
+  #define VECTOR __attribute__((vector_size(16)))
```
seiko2plus: why not `__vector`?

Configure.cmake
```diff
@@ -600,6 +636,46 @@ if (ENFORCE_VSX AND NOT COMPILER_SUPPORTS_VSX)
...
+option(ENFORCE_Z14 "Build fails if Z14 is not supported by the compiler" OFF)
+
+if(SLEEF_ARCH_S390X AND NOT DISABLE_Z14)
+  set (CMAKE_REQUIRED_FLAGS ${FLAGS_ENABLE_Z14})
+  CHECK_C_SOURCE_COMPILES("
+  #include <vecintrin.h>
+  #define VECTOR __attribute__((vector_size(16)))
+  int main() {
+    VECTOR double d;
+    d = __builtin_s390_vec_sqrt(d);
```
seiko2plus: Z13 and Z14 both have sqrt support for double-precision, I guess you should check single-precision instead.

src/arch/helpers390x_128.h
```diff
@@ -0,0 +1,535 @@
+//          Copyright Naoki Shibata 2010 - 2019.
```
seiko2plus: I think ZVECTOR can be combined with the VSX header, what do you think?

src/arch/helpers390x_128.h
```diff
@@ -0,0 +1,535 @@
...
+#define ACCURATE_SQRT
+#define FULL_FP_ROUNDING
+
+#include <vecintrin.h>
+
+#include <stdint.h>
+#include <math.h>
+#include "misc.h"
+
+#define VECTOR __attribute__((vector_size(16)))
```
seiko2plus: same here, why not using compiler token `__vector` instead?

src/arch/helpers390x_128.h
```diff
@@ -0,0 +1,535 @@
...
+
+typedef VECTOR unsigned long long vmask;
+typedef VECTOR unsigned long long vopmask;
+
+typedef VECTOR double vdouble;
+typedef VECTOR int vint;
+
+typedef VECTOR float vfloat;
```
seiko2plus: Z13 doesn't support single-precision, but it can be emulated via two double-precision registers

edelsohn:
z13 has special instructions to load and store a single precision floating point pair as a double precision floating point pair, and then use normal double precision operations.  z14 adds full single precision support.

edelsohn:
The Linux Community Cloud systems now are z15.

src/arch/helpers390x_128.h
shibatch: I think that would make the code confusing.

src/arch/helpers390x_128.h
shibatch: I don't understand this.
It seems that both gcc and qemu support single-precision vector operation with z13 option.
Are both of those buggy?

shibatch: Ah, gcc unrolls each vector operation into scalar operations.

Configure.cmake
shibatch: Done.

Configure.cmake
shibatch: Done.

Configure.cmake
shibatch: Done.

src/arch/helpers390x_128.h
shibatch: Done.

src/arch/helpers390x_128.h
seiko2plus: vector extension in both compilers(clang & gcc) always emulate non-existence instructions which may lead to bad performance, that why I prefer the use of the prototypes of `ZVECTOR` and `VSX` instead. 
An example of how we should handle single-precision in `Z13`:
```C

// type
#if CONFIG <= 131
typedef struct {
    vdouble val[2];
} vfloat;
#else
typedef __vector float vfloat;
#endif
// load
static INLINE vfloat vloadu_vf_p(const float *p)
{
#if CONFIG <= 131
    vfloat r;
    r.val[0] = vec_ld2f(p); // load and convert
    r.val[1] = vec_ld2f(p + 2);
    return r;
#else
	return vec_xl(0, p);
#endif
}
// store
static INLINE void vstoreu_v_p_vf(float *p, vfloat v)
{
#if CONFIG <= 131
    vec_st2f(v.val[0], p); // convert and store
    vec_st2f(v.val[1], p + 2);
#else
    return vec_xst(v, 0, p);
#endif
}
// Now emulate all operations via two double-prescion vectors
static INLINE vfloat vsqrt_vf_vf(vfloat vf)
{
#if CONFIG <= 131
    vf.val[0] = vec_sqrt(vf.val[0]);
    vf.val[1] = vec_sqrt(vf.val[1]);
    return vf;
#else
    return vec_sqrt(vf); 
#endif
}
````

Comment (shibatch) [reply src/arch/helpers390x_128.h]: This is emulation using double-precision operations.
SLEEF has functions that return bit-identical results across all platforms, and those functions cannot be implemented using this method. We need genuine single-precision operations.
I don't know how widely Z13 computers are currently deployed, but is single-precision support for Z13 so important?

**Comment by shibatch** [comment]
@seiko2plus @edelsohn 

I now wonder if it is worth implementing single-precision functions with emulation with double-precision vector computation with ZVECTOR1.
There are the following two obstacles in implementing the functions in that way.

* Data types for vfloat will be different between ZVECTOR1 and ZVECTOR2. This will make the code messy.
* We cannot implement the deterministic version of the functions. We need true single-precision computation for this.

Importance of such functions is not certain. I don't know how widely Z13 computers are being used. It is hard to imagine that users use such functions on only Z13 computers.

The reason that I implemented ZVECTOR1 support is that QEMU 4.2.0 supports up to Z13 processors. So, testing is possible for ZVECTOR1 without real hardware. I also did not notice that the single precision vector operations on Z13 are emulated within the compiler.

There are another option, which is to drop ZVECTOR1 support. So, there are three options.

1. Continue implementing single precision functions  with emulation with double-precision vector computation.
2. Go with the current implementation.
3. Drop ZVECTOR1 support.

I would like to know how important ZVECTOR1 support is. Double-precision functions with the current implementation works normally. We can just say that single precision functions with ZVECTOR1 are supplementary. So I think the current implementation is satisfactory. How do you guys think?

File: src/arch/helpers390x_128.h (hunk 6 above)
Comment (seiko2plus) [reply src/arch/helpers390x_128.h]: Z13 mainframe launched in 2015, not sure if still widely used but most of the Linux-one instances provided through IBM/Cloud are Z14 so I guess we can drop the support of Z13 for now,
and focuses only on Z14/Z15

**Comment by edelsohn** [comment]
The Linux Community Cloud systems all should be z15.
I think that you safely can ignore z13. z13 will not be important for PyTorch users.

File: travis/before_install.s390x-gcc.sh [hunk 7]
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+set -ev
```
seiko2plus: Could you please dump auxiliary vector via  `LD_SHOW_AUXV=1 /bin/true` to the ZARCH version?

shibatch: Done.

shibatch:
@seiko2plus @edelsohn 
I have now removed Z13 support.
Please review the patch again.

PR #292, opened 2020-03-24
Combined cleanup

shibatch:
This is a patch for updating and cleaning up, consisting of following points.

* Update copyright notice
* Update travis settings from xenial to bionic
* Update appveyor settings from VS2017 to VS2019
* Remove workaround for  cross-compilation for PPC64
* Add detection of child process crash to the tester


shibatch:
Okay, I will decompose this.

PR #295, opened 2020-03-25
Remove PPC64 workaround

shibatch:
This patch removes bug workarounds which were required when cross-compiling for PPC64.


fpetrogalli:
Are the 2 builds marked as `ppc64el` at [1] verifying that the changes i this patch are correct?

[1] https://travis-ci.org/github/shibatch/sleef/builds/666621212?utm_source=github_status&utm_medium=notification

shibatch:
Travis now has native PPC64 servers.
I also have access to a POWER8 VM.
So, cross-compilation is not needed anymore.

PR #300, opened 2020-04-02
Building with Clang on Windows

shibatch:
With this patch, the library can be built with Clang on Windows.


Configure.cmake
```diff
@@ -32,25 +25,34 @@ if (ENFORCE_TESTER3 AND NOT SLEEF_OPENSSL_FOUND)
...
-if (LIBFFTW3)
-  find_path(FFTW3_INCLUDE_DIR
-    NAMES fftw3.h
-    ONLY_CMAKE_FIND_ROOT_PATH)
-endif(LIBFFTW3)
-
-if (NOT LIBM)
-  set(LIBM "")
-endif()
+if (NOT SLEEF_CLANG_ON_WINDOWS)
```
fpetrogalli, line 28: I don't get this - why do you need to customize this for clang on windows? If the libraryes are not available, the variables below (`LIB_MPFR`, `LIBFFTW3`, ..) would evaluate to false and teh corresponding compilation flags will not be set. Is that breaking things?

Configure.cmake
```diff
@@ -310,6 +312,10 @@ if(CMAKE_C_COMPILER_ID MATCHES "(GNU|Clang)")
     string(CONCAT FLAGS_WALL ${FLAGS_WALL} " -Wno-psabi")
     set(FLAGS_ENABLE_NEON32 "-mfpu=neon")
   endif(CMAKE_C_COMPILER_ID MATCHES "GNU")
+
+  if (SLEEF_CLANG_ON_WINDOWS)
+    string(CONCAT FLAGS_WALL ${FLAGS_WALL} " -D_CRT_SECURE_NO_WARNINGS -Wno-deprecated-declarations")
```
fpetrogalli, line 317: Why is this needed?

shibatch: Because clang references MSVC header files.
They have deprecation and security attributes for many functions, and many deprecation warnings are displayed without these options.

Configure.cmake
shibatch: CMake finds libraries for Cygwin, and this causes build errors.
We need to check the builds with Cygwin testers.

fpetrogalli: May I ask you to add this explanation as a comment? It will help people like me who doesn't know much about compiling on windows in understanding why there is this `if not clang_on_windows` condition.

Configure.cmake
fpetrogalli: Please explain this with a comment in the sources too.

src/quad/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ foreach(SIMD ${SLEEFQUAD_SUPPORTED_EXT})
   endif()
 endforeach()
 
-if(MSVC OR MINGW AND WIN32)
+if(MSVC OR SLEEF_CLANG_ON_WINDOWS OR MINGW AND WIN32)
```
fpetrogalli, line 63: Why is this condition different from the one `if((MSVC OR MINGW AND WIN32) OR SLEEF_CLANG_ON_WINDOWS)` 25 of `src/libm/CMakeLists.txt`?

Configure.cmake
```diff
@@ -2,12 +2,9 @@ include(CheckCCompilerFlag)
...
 
-# Some toolchains require explicit linking of the libraries following.
-find_library(LIB_MPFR mpfr)
-find_library(LIBM m)
-find_library(LIBGMP gmp)
-find_library(LIBRT rt)
-find_library(LIBFFTW3 fftw3)
+# Detect if cmake is running on CI environments
+string(COMPARE NOTEQUAL "" "$ENV{TRAVIS}" RUNNING_ON_TRAVIS)
+string(COMPARE NOTEQUAL "" "$ENV{APPVEYOR}" RUNNING_ON_APPVEYOR)
```
fpetrogalli, line 7: Please make these variables to be set at `cmake` invocation, as in `cmake -DRUNNING_ON_TRAVIS=On ...`. It is safer than having to deal with things breaking in the case the name of the env variable change.

shibatch: Done.

src/quad/CMakeLists.txt
shibatch: Those two are now same.

Configure.cmake
shibatch: I added the comment.

PR #311, opened 2020-06-26
Fix remainder

shibatch:
This patch fixes a bug in remainder function.

### DIFF
--- a/src/libm-tester/tester.c
+++ b/src/libm-tester/tester.c
@@ -4348,6 +4348,7 @@ void do_test() {
     for(y = -1e+7;y < 1e+7 && success;y += 1.51e+5) {
       for(x = -1e+7;x < 1e+7 && success;x += 1.51e+5) checkAccuracy_f_f(mpfr_remainder, child_remainderf, y, x, 0.5);
     }
+    checkAccuracy_f_f(mpfr_remainder, child_remainderf, 11114942644092928, 224544296009728, 0.5);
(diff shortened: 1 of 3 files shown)


fpetrogalli:
Hi. Could you guard the fix with a test on the input value that was causing the failure?

shibatch:
Done.

src/libm-tester/tester.c
```diff
@@ -4348,6 +4348,7 @@ void do_test() {
     for(y = -1e+7;y < 1e+7 && success;y += 1.51e+5) {
       for(x = -1e+7;x < 1e+7 && success;x += 1.51e+5) checkAccuracy_f_f(mpfr_remainder, child_remainderf, y, x, 0.5);
     }
+    checkAccuracy_f_f(mpfr_remainder, child_remainderf, 11114942644092928.0, 224544296009728.0, 0.5);
```
fpetrogalli, line 4351: I think you want to add a test also for the double precision case.

shibatch: Only single precision cases are found.

fpetrogalli: I see - if that's the case, I think you should revert the change in the double precision implementation.

shibatch: No. Those two implementations work with the same principle. So, if the single-precision version fails, there should be cases for the double-precision version to fail.

PR #312, opened 2020-07-05
Correctly preserve default CFLAGS

yuyichao:
1. `string(CONCAT)` does not append content to the string but replaces the content of the output variable. Appending should use `string(APPEND)` instead.
2. The contents of the two arguments should be concatenated with a space so
   `string(APPEND)`/`string(CONCAT)` aren't the right/most convinient functions to use either.
   Since there're only two strings input, simply use a string interpolation instead.


shibatch:
Thank you for your patch. It is now merged.

PR #314, opened 2020-07-19
Include version numbers in public header.

nemequ:
This can be used to detect whether the version of SLEEF being used
supports a feature.  For example

  #if SLEEF_VERSION_MAJOR > 3 || \
      (SLEEF_VERSION_MAJOR == 3 && SLEEF_VERSION_MINOR >= 5)
    /* Use function only available in SLEEF 3.5+ */
  #endif


shibatch:
Thank you for your patch.

PR #315, opened 2020-07-22
[AltiVec] include altivec.h when __ALTIVEC__ is defined.

nemequ:
sleef.h will use `vector ...` types unprotected on POWER, which
requires including AltiVec just like, for example NEON requires
arm_neon.h.  However, unlike NEON, SVE, and SSE, until this patch
SLEEF required including altivec.h prior to including sleef.h.

I'm not all that familiar with SLEEF yet, so this is basically the smallest change I could make to get the effect I was after and matches the behavior of the other ISA extensions.  However, I think this is actually the wrong way to go about this.  SLEEF conditionally includes these headers and then proceeds to declare functions which depend on them unconditionally.

On POWER I think this is actually safe for relatively recent revisions of the ISA, but it's pretty common to compiler on Arm without NEON support, and while on x86_64 you'll always have at least SSE2, on x86 that isn't so.  In these cases, even if SLEEF was compiled with support for the relevant ISA extensions, that's not necessarily true for code using the public API.

I'm not quite sure where the architecture-specific function definitions are coming from, but perhaps it would be better to simply unconditionally include the relevant headers there, and drop the conditional includes from the `sleeflibm_header.h.orig`?

If that sounds better to you I'd be happy to try and re-work the patch to do that instead.

By the way, just so you have a bit of context for my patches, I'm working on integrating SLEEF into [SIMDe](https://github.com/simd-everywhere/simde), particularly our SVML implementation (which is currently mostly libm in a loop, which is obviously not optimal).


src/libm/sleeflibm_header.h.org
```diff
@@ -54,6 +54,10 @@
 #include <arm_sve.h>
 #endif
 
+#if defined(__ALTIVEC__)
+#include <altivec.h>
+#endif
```
seiko2plus, line 59: Currently, Sleef supports VSX starting from (ISA2.07) with little-endian mode only.
```suggestion
#if defined(__VSX__) &&  defined(__PPC64__) && defined(__LITTLE_ENDIAN__)
    #include <altivec.h>
    #ifdef __cplusplus
        #undef bool
        #undef vector
        #undef pixel
    #endif
#endif
```

nemequ: ISA 2.07 makes sense; IIRC that's when support for 64-bit values was added to AltiVec/VSX.

However, what you suggested doesn't actually test for ISA 2.07; `__VSX__` is defined on ISA 2.06 (POWER 7) as well. I think what you really want is `#if defined(__VSX__) && defined(__ARCH_PWR8) && defined(__LITTLE_ENDIAN__)`.  I'll update the patch with that, but I still wonder if this is really the right place for it.

As for undefining `bool`/`vector`/`pixel`, that actually causes compiler errors for me.  For example:

```
[1/746] Compiling C++ object test/x86/mmx-native-cpp.p/meson-generated_.._mmx.cpp.o
FAILED: test/x86/mmx-native-cpp.p/meson-generated_.._mmx.cpp.o 
c++ -Itest/x86/mmx-native-cpp.p -Itest/x86 -I../test/x86 -I. -I.. -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -g -mcpu=native -fopenmp-simd -DSIMDE_MATH_SLEEF_ENABLE -DSIMDE_ENABLE_OPENMP -Wno-psabi -DSIMDE_TEST_BARE -MD -MQ test/x86/mmx-native-cpp.p/meson-generated_.._mmx.cpp.o -MF test/x86/mmx-native-cpp.p/meson-generated_.._mmx.cpp.o.d -o test/x86/mmx-native-cpp.p/meson-generated_.._mmx.cpp.o -c test/x86/mmx.cpp
In file included from ../test/x86/../../simde/simde-math.h:67,
                 from ../test/x86/../../simde/simde-common.h:44,
                 from ../test/x86/../test.h:6,
                 from ../test/x86/test-x86.h:4,
                 from ../test/x86/test-mmx.h:4,
                 from test/x86/mmx.cpp:25:
/usr/include/powerpc64le-linux-gnu/sleef.h:289:3: error: ‘vector’ does not name a type; did you mean ‘vec_or’?
  289 |   vector double x, y;
      |   ^~~~~~
      |   vec_or
```

This is with GCC 10.1.0 on PPC64LE Debian.

I understand what undefining them is supposed to do and I'm not opposed to it in priciple, but I think we would first need to figure out under exactly which circumstances it is safe to do so.

nemequ: I updated it as described abovo, but after thinking about this a bit more I'm not sure this makes sense either.

The x86 intrinsics are included when targeting x86/x86_64, or whenever MSVC is being used (it does target ARM, too), regardless of whether SSE is supported (which it always is on x86_64, but may not be on x86).

`arm_neon.h` is included when NEON support is enabled, but it's possible people will be compiling with NEON disabled for most of their code and still want to be able to call the NEON SLEEF functions if they determine at runtime that the CPU supports NEON.  Similar situation for SVE.  Right now, in those cases there will be a compile-time error because SLEEF uses the vector types unconditionally later in the file.

I think that really the only things that would make sense would be to include x86intrin.h/intrin.h (or just immintrin.h) on x86, arm_neon.h on ARMv7/ARMv8/AArch64, arm_sve.h on AArch64, and altivec.h on POWER (or maybe little-endian POWER; I don't think you can mix LE/BE in the same program, but I'm not certain).

Or make the includes part of the platform specific code instead of putting them in the common header.

PR #319, opened 2020-07-31
Add pkg-config support

nemequ:
I also noticed that the subdirectories were hardcoded, so I fixed that, too.

This only adds a pkg-config file for libsleef, but if you want I can add files for the other libraries as well, just let me know.


nemequ:
I rebased this, but I have no idea why the Windows builds are failing and don't have a way to reproduce, so I'm going to have to leave that up to someone else, sorry :(

PR #322, opened 2020-08-19
Build static bins if BUILD_SHARED_LIBS is defined

steplong:
Search for the static libraries instead and add -static to CFLAGS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 # Options
 
 option(BUILD_SHARED_LIBS "Build shared libs" ON)
+option(BUILD_STATIC_TEST_BINS "Build statically linked test executables" OFF)
 option(ENABLE_LTO "Enable LTO on GCC or ThinLTO on clang" OFF)
 option(BUILD_LIBM "libsleef will be built." ON)
 option(BUILD_DFT "libsleefdft will be built." ON)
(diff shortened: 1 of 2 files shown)


steplong:
@fpetrogalli it looked like BUILD_SHARED_LIBS only builds the libraries statically. Maybe, I should add a new variable like BUILD_SHARED_BINS instead of using BUILD_SHARED_LIBS. Also, it looked like the tests were failing silently when they ran suspiciously fast. If I passed -VV to ctest, it showed that it wasn't passing the interpreter to QEMU.

shibatch:
I don't understand the purpose well.
I think adding -static should be done automatically by cmake when BUILD_SHARED_LIBS=FALSE?

steplong:
When I set BUILD_SHARED_LIBS=FALSE, it only determined whether a static archive or shared object was built. The test executables that were going to be ran with QEMU were still dynamically linked.

shibatch:
Okay, but why do you need static executables?

steplong:
On some of the machines we will run the tests on, we might not have an elf interpreter to pass to QEMU.

shibatch:
You see, builds on mac are failing with the current patch.
Can you fix this problem?

shibatch:
Please refer to the following page.

https://stackoverflow.com/questions/24648357/compiling-a-static-executable-with-cmake

steplong:
Hmmm, for the appveyor build, it looks like it can't find OpenSSL. Also, I can't seem to see the details for the jenkins jobs. I get a 404 when I click the details link.

shibatch:
Maybe it is better to add a separate flag to enable this feature.

steplong:
That makes sense. I'll add BUILD_STATIC_TEST_BINS then.

PR #342, opened 2020-10-05
Fix build problem with gcc-10 on ppc64

shibatch:
This is caused because __float128 is defined since gcc-8, and even though it does not define `__SIZEOF_FLOAT128__`.


kif:
Indeed, it works now. Thanks !

PR #348, opened 2020-10-12
Add RUNTIME destination to the install() commands for libraries.

bluescarni:
Hello,

I am trying to create a SLEEF package for conda, and I noticed that the CMake installation command for Windows will install only the ``.lib`` files but not the ``.dll`` files:

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=222443&view=logs&j=240f1fee-52bc-5498-a14a-8361bde76ba0&t=23010c01-6a9d-5f94-9037-792618644b77&l=3695

I believe this is due to the fact that the CMake ``install()`` commands for the libraries are missing the ``RUNTIME DESTINATION`` arguments - DLLs are normally installed in the same place as executables in Windows.

I have opened this PR with a tentative fix, but I am not able to test it on a Windows machine at this time.


shibatch:
Merged. Thank you for your contribution.

PR #351, opened 2020-10-26
Fix tester3 for aarch64 with SVE

zatrazz:
The tester3 file is included by multiple files and built with different
compiler options.  The __ARM_FEATURE_SVE is defined by the compiler iff
SVE is explicit added on -march (for instance, armv8.2-a+sve).  Building
with gcc, it fails with:

[ 30%] Building C object src/libm-tester/CMakeFiles/tester3advsimdnofma.dir/tester3.c.o
/home/azanella/projects/sleef/src/libm-tester/tester3.c: In function ‘do_test’:
/home/azanella/projects/sleef/src/libm-tester/tester3.c:183:43: error: incompatible type for argument 1 of ‘vd2getx_vd_vd2’
  183 |     double fxx = GET(TYPE)(vd2getx_vd_vd2(vx2), r), fxy = GET(TYPE)(vd2gety_vd_vd2(vx2), r); \
      |                                           ^~~
      |                                           |
      |                                           Sleef_float64x2_t_2
/home/azanella/projects/sleef/src/libm-tester/tester3.c:195:7: note: in expansion of macro ‘exec_d2_d’
  195 |       exec_d2_d(ATR, NAME, ULP, DPTYPE, DPTYPESPEC, EXTSPEC, d);        \
      |       ^~~~~~~~~
/home/azanella/projects/sleef/src/libm-tester/tester3.c:317:3: note: in expansion of macro ‘test_d2_d’
  317 |   test_d2_d(sincos, u10, -1e+14, 1e+14, 200001);
      |   ^~~~~~~~~
/home/azanella/projects/sleef/src/libm-tester/tester3.c:84:49: note: expected ‘svfloat64x2_t’ but argument is of type ‘Sleef_float64x2_t_2’
   84 | static svfloat64_t vd2getx_vd_vd2(svfloat64x2_t v) { return svget2_f64(v, 0); }
      |                                   ~~~~~~~~~~~~~~^

Because __ARM_FEATURE_SVE will defined even when the testing is
building for a different vector ABI.

Instead of using __ARM_FEATURE_SVE, check for specifically define
AARCH64_SVE instead.


shibatch:
@PierreBlanchard @fpetrogalli I want some person at Arm to review this patch.

shibatch:
@zatrazz I don't understand why you need such a feature.
SVE is automatically detected by the build system.
Please explain in what situation you need to explicitly add -march option.

zatrazz:
> @zatrazz I don't understand why you need such a feature.
> SVE is automatically detected by the build system.
> Please explain in what situation you need to explicitly add -march option.

It come from my misunderstanding on how the SVE supported was added, sorry for the noise. However it might still fail if gcc is build to set SVE as the default ABI (although not sure how you want to handle it since it seems an unlikely scenario).

shibatch:
I still don't understand, but if you show some specific usage scenario, I might be able to improve your situation.

PR #357, opened 2020-11-24
Fix undefined behavior in `xnextafterf`

malfet:
C standard does not define how sign bit should be handled during left shift, which triggers UBSAN runtime error:
```
src/libm/sleefsimdsp.c:3031:101: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
```
Discovered while working on https://github.com/pytorch/pytorch/pull/48275


shibatch:
SLEEF only supports architectures with two's-complement number system.
It strongly depends on two's-complement number system.

Issue #313, opened 2020-07-17
POWER VSX PowerPC64LE Linux support


edelsohn:
Port Sleef to IBM POWER VSX architecture supporting PowerPC64LE Linux compiled with both GCC and Clang. Achieve equivalent speedup to x86-64 and AArch64 appropriate for the VSX 128 bit vector width.

With Github clone from today on CentOS8 built with GCC AT 12.0, I see the following failures:
The following tests FAILED:
	 36 - roundtriptest2dsp_4_4 (Failed)
	 37 - roundtriptest2dsp_8_8 (Failed)
	 38 - roundtriptest2dsp_10_10 (Failed)
	 39 - roundtriptest2dsp_5_15 (Failed)

seiko2plus:
> IBM POWER VSX architecture supporting PowerPC64LE Linux compiled with both GCC and Clang

it's already ported and It was tested on wide versions of GCC and CLANG during working on gh-288

> Achieve equivalent speedup to x86-64 and AArch64 appropriate for the VSX 128 bit vector width.

That will require a full review of all existing kernels. I think it's better to merge #291(still need improvements similar to what happened on gh-288) and then review both VSX/zVX in the same time.

> With Github clone from today on CentOS8 built with GCC AT 12.0, I see the following failures:

please could you set `export CTEST_OUTPUT_ON_FAILURE=TRUE` before calling `ctest` to expose the error messages?

edelsohn:
Start 36: roundtriptest2dsp_4_4
36/39 Test #36: roundtriptest2dsp_4_4 ............***Failed   64.36 sec
Path(random) :3(ST) 1(ST) 
ISA : VSX 128 bit float
transpose NoMT(measured): 18470
transpose   MT(measured): 64334599
Path(random) :2(ST) 2(ST) 
ISA : VSX 128 bit float
transpose NoMT(loaded): 18470
transpose   MT(loaded): 64334599
complex : NG (1.86187)

      Start 37: roundtriptest2dsp_8_8
37/39 Test #37: roundtriptest2dsp_8_8 ............***Failed    0.55 sec
Path(random) :4(ST) 2(ST) 1(ST) 1(ST) 
ISA : VSX 128 bit float
transpose NoMT(measured): 17325
transpose   MT(measured): 497714
Path(random) :2(ST) 2(ST) 3(ST) 1(ST) 
ISA : VSX 128 bit float
transpose NoMT(loaded): 17325
transpose   MT(loaded): 497714
[...]

PR #360, opened 2020-11-28
Add power9 support

shibatch:
This patch adds POWER9 support as mentioned in issue https://github.com/shibatch/sleef/issues/313.


RajalakshmiSR:
Thanks @shibatch for adding this. I am able to see -mcpu=power9 in the build log on POWER9. 
However, when I run 'make test' with this patch on POWER8 system, I could still see the new vsx3 tests in log. Is this expected?

shibatch:
@RajalakshmiSR Yes. If the IUT successfully detects that the corresponding vector extension is not available, it exits with SUCCESS code, displaying the following message.

```
[sleef-ppc64]~/work/sleef3/sleef-power9/build-ppc64le/bin$ ./iutvsx3


***** This host does not support the necessary CPU features to execute this program *****


0
[sleef-ppc64]~/work/sleef3/sleef-power9/build-ppc64le/bin$ ./tester ./iutvsx3


*** Now testing ./iutvsx3


***** This host does not support the necessary CPU features to execute this program *****




Tester : *** CPU does not support the necessary feature
[sleef-ppc64]~/work/sleef3/sleef-power9/build-ppc64le/bin$
```

RajalakshmiSR:
@shibatch There are some ifdefs for clang and gcc in helperpower_128.h. 
Gcc 7 supports vec_xxpermdi , __builtin_vsx_xvcvsxwdp, __builtin_vsx_xvcvdpsxws. If we expect minimum gcc version as 7, then we can remove the differences between clang and gcc. (Tried with ubuntu gcc 7.5)
Also __builtin_convertvector and  vec_neg are available in GCC9 . So when we add minimum gcc version as 9, we can just make that code common for both clang and gcc.

shibatch:
Those built-in functions are now used according to the gcc version by checking __GNUC__ macro.
I manually ran the tests on gcc-6, gcc-7 and gcc-9, and they all worked.

RajalakshmiSR:
LGTM. Thanks @shibatch

shibatch:
Merged. Thank you for reviewing.

PR #392, opened 2021-01-13
[Cleanup] Conform strict aliasing rules

shibatch:
This is a combined patch for removing potential problems with the strict aliasing rule.
It also drops long double support for DFT.


shibatch:
The basic way of conforming to the strict aliasing rule in this patch is to stop using unions.
There were many places where reinterpreting cast by a union is used. These are substituted with memcpy in this patch.
For example,
```
  union {
    double f;
    int64_t i;
  } tmp;
  tmp.f = d;
  return tmp.i;
```
is substituted with the following code.
```
  int64_t ret;
  memcpy(&ret, &d, sizeof(ret));
  return ret;
```
This is necessary because the code is sometimes compiled as a C++ source code.

Issue #417, opened 2021-04-04
Issues with older CMake (3.11) after last commit


zachmatson:
In CMake 3.11, I was previously able to build SLEEF. After the latest commit (eb3d977) I get the error
```bash
CMake Error at src/libm/CMakeLists.txt:1066 (install):
    install TARGETS given no ARCHIVE DESTINATION for static library target
    "sleef".
```

shibatch:
Please check if this patch solves your problem.

https://github.com/shibatch/sleef/tree/fix-static-with-old-cmake

zachmatson:
That fixes it for me

friendlyanon:
`ARCHIVE` is already provided: https://github.com/shibatch/sleef/blob/eb3d97785cb4f6f6f4c0ffd6570ca6c73ccac2ec/src/libm/CMakeLists.txt#L1057-L1059

Please describe your build environment, because this is not an issue with CMake nor the changes. See these [CMake 3.8 CI runs](https://github.com/friendlyanon/cxx-static-shared-example/runs/2180308366?check_suite_focus=true) with the same install commands.

PR #419, opened 2021-04-05
Provide dummy destination in namelink install

friendlyanon:
> A DESTINATION without a block will provide a default for all blocks,
> because DESTINATION is always required prior to CMake 3.14, even if it's
> already provided in a previous install rule and install rules for
> targets had to be split in two prior to CMake 3.12.

Fixes: #417


friendlyanon:
The failures are unrelated Clang license issues.

PR #454, opened 2023-03-08
Rename global variable to avoid duplicate definitions

nisanthmp:
The vairable sleef_cpuidtmp2 is already defined in the common code (src/libm/dispscalar.c.org) in the global scope.
Its redefinition in Power specific code (src/libm/tryvsx3.c) in global scope was causing compilation error on Power10. It is fixed now by renaming the variable in Power specific code.

There are test failures still present on Power10, that need to be fixed.

### DIFF
--- a/src/libm/tryvsx3.c
+++ b/src/libm/tryvsx3.c
@@ -1,8 +1,8 @@
 #include <altivec.h>
 
 __vector double sleef_cpuidtmp0;
-__vector unsigned long long sleef_cpuidtmp1, sleef_cpuidtmp2;
+__vector unsigned long long sleef_cpuidtmp1, sleef_cpuidtmp3;
 
 void sleef_tryVSX3() {
-  sleef_cpuidtmp0 = vec_insert_exp(sleef_cpuidtmp1, sleef_cpuidtmp2);
+  sleef_cpuidtmp0 = vec_insert_exp(sleef_cpuidtmp1, sleef_cpuidtmp3);
 }


ladybug13:
Hi, 

We are currently looking into this.

We recently refactored and restarted our CI and we cannot observe any test failures for Power10 architecture.

That being said, the current tests in place for Power10 result from cross-compilation and emulation in QEMU, which means it is not necessarily representative of your environment.

Therefore, it would be helpful to us if you could you provide some further information and logs so it makes more clear the issue you were having and how it is being fixed. The information we are looking for is:

* Your environment, namely the **OS**, the **architecture** and **machine**, the **compiler** and **compiler version**, the **cmake version** as well as the full **cmake command** you are using when building SLEEF.
* The tests that are failing. Adding the following flags to the cmake command when building sleef: `-DENFORCE_TESTER=ON -DENFORCE_TESTER3=ON` will provide more information on the failures.

Furthermore, before we can merge your fix could you rebase your commit in order to trigger the new CI on these changes?

Thank you so much for the contribution.

joanaxcruz:
Hi,

We are currently looking into this.

We recently refactored and restarted our CI and we cannot observe any test failures for Power10 architecture.

That being said, the current tests in place for Power10 result from cross-compilation and emulation in QEMU, which means it is not necessarily representative of your environment.

Therefore, it would be helpful to us if you could you provide some further information and logs so it makes more clear the issue you were having and how it is being fixed. The information we are looking for is:

Your environment, namely the **OS**, the **architecture** and **machine**, the **compiler** and **compiler version**, the **cmake version** as well as the full **cmake command** you are using when building SLEEF.
The tests that are failing. Adding the following flags to the cmake command when building sleef: `-DENFORCE_TESTER=ON -DENFORCE_TESTER3=ON` will provide more information on the failures.
Furthermore, before we can merge your fix could you rebase your commit in order to trigger the new CI on these changes?

Thank you so much for the contribution.

blapie:
Closing this as should be fixed by #499. Feel free to re-open if problem still appears.

PR #476, opened 2023-11-03
Add CI on Github Actions for x86_64, aarch64, arm, ppc64 and s390x

luhenry:
As discussed offline, this PR adds some CI on GitHub Actions. That's expected to facilitate contribution by having smoke testing on any changes that would be integrating, facilitating the work of maintainers by making sure a change will not completely break all supported architectures (aarch64, arm, ppc64, s390x, and x86_64).

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,0 +1,117 @@
+
+name: "Build & Test"
+
+on:
+  # allow direct trigger
+  workflow_dispatch:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4.1.1
+        with:
+          persist-credentials: false
+
+      - name: Install dependencies
+        run: sudo apt-get update -y -qq && sudo apt-get install -y -qq build-essential clang curl ninja-build libgmp-dev libmpfr-dev cpuinfo
+
+      - name: CPU Info
+        run: isa-info
+
+      - name: Build native
+        run: |
+          cmake -S . -B _build-native -GNinja \
+            -DCMAKE_INSTALL_PREFIX=$(pwd)/_install-native \
+            -DBUILD_STATIC_TEST_BINS=ON
+          cmake --build _build-native
+          cmake --install _build-native
+
+      - name: Test native
+        run: |
+          cd _build-native
+          ctest -j$(nproc)
+
+  build-and-test-cross:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # AArch64
+          - arch: aarch64
+            qemu_cpu: "max,sve=off"
+          - arch: aarch64
+            qemu_cpu: "max,sve=on,sve128=on"
+          - arch: aarch64
+            qemu_cpu: "max,sve=on,sve256=on"
+          - arch: aarch64
+            qemu_cpu: "max,sve=on,sve512=on"
+          # Aarch32
+          - arch: armhf
+            binfmt: arm
+            package: -arm-linux-gnueabihf
+          # PPC64
+          - arch: ppc64el
+            binfmt: ppc64le
+            package: -powerpc64le-linux-gnu
+          # IBM Z
+          - arch: s390x
+
+    steps:
+      - uses: actions/checkout@v4.1.1
+        with:
+          persist-credentials: false
+
+      - uses: docker/setup-qemu-action@v3.0.0
+        with:
+          platforms: ${{ matrix.binfmt || matrix.arch }}
+
+      - name: Install dependencies
+        run: sudo apt-get update -y -qq && sudo apt-get install -y -qq build-essential clang curl ninja-build libgmp-dev libmpfr-dev gcc${{ matrix.package || format('-{0}-linux-gnu', matrix.arch) }}
+
+      - name: Print host CPU info
+        run: |
+          cat /proc/cpuinfo
+
+      - name: Build native
+        run: |
+          cmake -S . -B _build-native -GNinja \
+            -DCMAKE_INSTALL_PREFIX=$(pwd)/_install-native \
+            -DBUILD_STATIC_TEST_BINS=ON \
+            -DENFORCE_TESTER=ON
+          cmake --build _build-native
+          cmake --install _build-native
+
+      - name: Build ${{ matrix.arch }}
+        env:
+          CTEST_OUTPUT_ON_FAILURE: "TRUE"
+          QEMU_CPU: ${{ matrix.qemu_cpu }}
+        run: |
+          cmake -S . -B _build-${{ matrix.arch }} -GNinja \
+            -DCMAKE_TOOLCHAIN_FILE=$(pwd)/travis/toolchain-${{ matrix.arch }}.cmake \
+            -DCMAKE_INSTALL_PREFIX="$(pwd)/_install-${{ matrix.arch }}" \
+            -DNATIVE_BUILD_DIR="$(pwd)/_build-native" \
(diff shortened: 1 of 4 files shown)


luhenry:
Some known limitations:
1. build instability with `bin/testervecabi` sometimes failing to build. That's fixed by hitting "Re-run failed jobs" enough times
2. no tests are run for arm, s390x, and ppc64. `qemu_cpu` needs to be specified to enable the correct architecture-specific extensions (ex: vsx for s390x). I suggest we do that in a follow-up PR so that I can focus on integrating the RISC-V support as discussed off-line.

luhenry:
The GitHub Actions run is available at https://github.com/rivosinc/sleef/actions/runs/6745836169

blapie:
Thank you very much for this PR, this is very helpful! 

> build instability with bin/testervecabi sometimes failing to build. That's fixed by hitting "Re-run failed jobs" enough times

Cannot wrap my head around these random failures. Cannot reproduce locally, but my environment differs slightly.
What jobs/steps were failing? Was it always the native build?
There is quite a number of options that were passed to cmake in the original Jenkins jobs, see [Jenkinsfile](https://github.com/shibatch/sleef/blob/master/Jenkinsfile) for x86 with gcc or clang.
These probably were hiding the bug you are seeing. Do you get similar failures when matching options in Jenkinsfile:
```
-DSLEEF_SHOW_CONFIG=1 -DENFORCE_TESTER3=TRUE -DBUILD_SCALAR_LIB=TRUE \
-DBUILD_QUAD=TRUE -DBUILD_DFT=TRUE -DBUILD_INLINE_HEADERS=TRUE -DBUILD_SHARED_LIBS=FALSE \
-DDISABLE_FMA4=TRUE -DENFORCE_SSE2=TRUE -DENFORCE_SSE4=TRUE \
-DENFORCE_AVX=TRUE -DENFORCE_AVX2=TRUE -DENFORCE_AVX512F=TRUE 
```

for gcc Jenkins also has `-DENABLE_CXX=TRUE `.

Have you tried to set the number of threads to 1 in `cmake --build`, although I'd be surprised if ninja failed to build in parallel?

>  I suggest we do that in a follow-up PR so that I can focus on integrating the RISC-V support as discussed off-line.

Absolutely, this is still very minimalistic in terms of cmake configuration as well. But it is a great start to unblock pending PRs.

luhenry:
The issue with linking seems to be in name variation for the symbols. Let's look at `Sleef_tanhf_u10` for example. In a functioning build, the symbol name will be `_ZGVdN8v_Sleef_tanhf_u10` while in the failing build it will be `_ZGVeN8v_Sleef_tanhf_u10` (notive the `_ZGVd` vs `_ZGVe`). The compilation commands and compiler versions are exactly the same, so I don't know where the variation comes from.

luhenry:
Issue seems to be with how `testervecabi.c.o` is compiled and various generation of GitHub runners. Given `testervecabi.c.o` is compiled with `-march=native`, different generation of GitHub runners will have different CPU features. That will lead to generating different mangled names for the symbols, while the libsleef.a object files have not been compiled with the same `-march=native`.

Fix would likely be to replace `-march=native` by something more deterministic.

luhenry:
That's the latest run without `-march=native` for `testervecabi` https://github.com/rivosinc/sleef/actions/runs/6754112260, everything passes on first attemp.

luhenry:
I've enabled testing on QEMU for ppc64 and arm. I couldn't figure out the magic combination for s390x just yet. I'll need some more domain-specific expertise on s390x to enable the right options, I don't understand what they are.

blapie:
Hi! Very good point. I would like to make sure we are not losing the point of testvecabi by doing this, `-march-native` was introduced at the same time as #404, which makes me think it is a feature. 
Is there a way make sure the library and tests are built with the same generation of runners (possibly prescribe what generation we intend to test on)? Alternatively could this be done in a single job?

We'll take over for s390x, this already provides great coverage, thank you!

luhenry:
> Is there a way make sure the library and tests are built with the same generation of runners (possibly prescribe what generation we intend to test on)? Alternatively could this be done in a single job?

It is already built all on the same runner, in the same job. The issue is one is compiled without `-mavx512f` and the other is (via `-march=native`). So it seems indeed that it's an issue with the ABI uncovered but unrelated to this PR.

blapie:
Right, got you! Thanks, we can look at this outside this PR. 
It is only a fairly recent tester, I wonder if that ever was tested extensively.
Just opened an issue #478 for this.

luhenry:
@blapie are you happy to merge this PR as-is (with the removal of `-march=native`) and merge it back once it's been fixed? Would there be anything else you want to get fixed/changed before we merge? Thank you!

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,117 @@
...
+          cmake -S . -B _build-${{ matrix.arch }} -GNinja \
+            -DCMAKE_TOOLCHAIN_FILE=$(pwd)/travis/toolchain-${{ matrix.arch }}.cmake \
+            -DCMAKE_INSTALL_PREFIX="$(pwd)/_install-${{ matrix.arch }}" \
+            -DNATIVE_BUILD_DIR="$(pwd)/_build-native" \
+            -DSLEEF_SHOW_CONFIG=1 \
+            -DDISABLE_SSL=ON \
+            -DBUILD_DFT=OFF \
+            -DBUILD_GNUABI_LIBS=OFF \
+            -DBUILD_INLINE_HEADERS=OFF \
+            -DBUILD_QUAD=OFF \
```
blapie, line 117: Could we please re-enable DFT, QUAD and INLINE_HEADERS? Is there a particular reason to disable GNUABI_LIBS?

blapie:
> are you happy to merge this PR as-is (with the removal of -march=native) and merge it back once it's been fixed?

Yes, I think that can be merged back later. I left a few comments on your patch.

luhenry:
@blapie I've updated the PR per your review, and I've split off the native build to reuse for the cross builds. You can find a run at https://github.com/rivosinc/sleef/actions/runs/6790080413

However, I'm running into an issue on `armhf` specifically as it fails to find the `vdouble_neon32_sleef` type:
```
/workspace/sleef/_build-armhf/include/sleefinline_neon32.h:3915:3: error: unknown type name ‘vfloat_neon32_sleef’
 3915 |   vfloat_neon32_sleef x, y;
```

The issue seems to be that there is no definition of `vdouble` in `src/arch/helperneon32.h`, which I expect is because `float64x2_t` is no defined on 32bits platforms.

I don't know how to fix it, but it's making the PR fails as it is.

blapie:
@luhenry Thanks for the update and further investigation! It looks like aarch32 might require a different environment or a fix. Do test pass when disabling inline headers for armhf? If so, I'll merge as is.

.github/workflows/build_and_test.yml
```diff
@@ -164,11 +169,20 @@ jobs:
...
           - arch: aarch64
-            qemu_cpu: "max,sve-max-vq=1"
+            qemu_cpu: "max,sve=off"
+          - arch: aarch64
+            qemu_cpu: "max,sve=on,sve128=on"
+          - arch: aarch64
+            qemu_cpu: "max,sve=on,sve256=on"
+          - arch: aarch64
+            qemu_cpu: "max,sve=on,sve512=on"
+          - arch: aarch64
```
blapie, line 182: I think we can spare some time and energy by removing 1024 and 2048.

blapie (changes requested):

blapie:
Create #480 issue to track this failure.
If you're happy with your current version (minus 1024 and 2048 vl) I'm happy to merge.

luhenry:
I've removed 1024 and 2048 bits. Last run at https://github.com/rivosinc/sleef/actions/runs/6813866461. Happy to merge as-is! Thank you again

blapie (approved):

PR #449, opened 2023-01-13
Add support for the RISC-V Vector ISA

ericlove:
Hello,

SiFive would like to contribute our port of SLEEF to the [RISC-V Vector architecture](https://github.com/riscv/riscv-v-spec) ("RVV").  This PR adds support for two new extensions, `RVVM1` and `RVVM2`, corresponding to RISC-V vector types with [LMUL](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#342-vector-register-grouping-vlmul20) set to 1 and 2 respectively (for further rationale behind those choices, see our comments in [helperrvv.h](https://github.com/sifive/sifive-sleef/blob/3d16b2cde43b8dfb54ddec02b5b66c221d2b36ee/src/arch/helperrvv.h#L70)).

The resulting `sleef.h` for this API looks roughly as follows:
```
SLEEF_IMPORT SLEEF_CONST vfloat64m1_t Sleef_sindx_u35m1(vfloat64m1_t);
SLEEF_IMPORT SLEEF_CONST vfloat64m1_t Sleef_finz_sindx_u35m1(vfloat64m1_t);
// [...]
SLEEF_IMPORT SLEEF_CONST vfloat32m1_t Sleef_sinfx_u35m1(vfloat32m1_t);
SLEEF_IMPORT SLEEF_CONST vfloat32m1_t Sleef_finz_sinfx_u35m1(vfloat32m1_t);
// [...]
SLEEF_IMPORT SLEEF_CONST vfloat64m2_t Sleef_sindx_u35m2(vfloat64m2_t);
SLEEF_IMPORT SLEEF_CONST vfloat64m2_t Sleef_finz_sindx_u35m2(vfloat64m2_t);
// [...]
SLEEF_IMPORT SLEEF_CONST vfloat32m2_t Sleef_sinfx_u35m2(vfloat32m2_t);
SLEEF_IMPORT SLEEF_CONST vfloat32m2_t Sleef_finz_sinfx_u35m2(vfloat32m2_t);
// [...]
```

__Building and Testing__
[...]

PR #448, opened 2022-10-19
Extend FMA support to RISC-V

GlassOfWhiskey:
At the moment, SLEEF fails to compile on the RISC-V architecture. The problem is that the RISC-V architecture is not listed among those supporting Fused Multiply-Add. This PR just adds the `__riscv` flag to the list of architectures supporting FMA.

PR #477, opened 2023-11-04
Integrate RISC-V support

luhenry:
It integrates the following changes:
- https://github.com/shibatch/sleef/pull/448
- https://github.com/shibatch/sleef/pull/449
- adding CI on GitHub Actions for linux-riscv64


luhenry:
This PR depends on https://github.com/shibatch/sleef/pull/476

blapie (commented):
What is the rationale for adding the prefix ?

blapie:
Hi!
Could you please put a link to passing GHA workflow?
Is there a chance this can be tested with a clang/llvm toolchain as well? SLEEF did not tend to merge features if not supported in both?

luhenry:
> What is the rationale for adding the prefix ?

The intrinsics' name changed upstream. It's purely to follow with the spec and not a design preference.

>  Could you please put a link to passing GHA workflow?

https://github.com/rivosinc/sleef/actions/runs/6754458819

> Is there a chance this can be tested with a clang/llvm toolchain as well? SLEEF did not tend to merge features if not supported in both?

It currently only works with clang as gcc doesn't have support for the RISC-V RVV intrinsics just yet.

blapie:
> It currently only works with clang as gcc doesn't have support for the RISC-V RVV intrinsics just yet.
Got you. I understand that it might take some time to be supported in gcc. It would be a shame not to make this available though.

blapie:
@ericlove @GlassOfWhiskey Would you be able to review the extra changes on top of yours?
@luhenry Do you know someone else the community that could review?

luhenry:
@blapie I've rebased the PR on top of the initial CI.

GlassOfWhiskey:
Hi all,
I'll review this today or tomorrow. Sry for the late reply.

GlassOfWhiskey:
As a first comment, since the RISC-V vector intrinsics standard has not reached a stable version uet, it could be useful to document which [version of the intrinsics](https://github.com/riscv-non-isa/rvv-intrinsic-doc) are we using right now.

luhenry:
I’ll also look into adding gcc + llvm compilation as both of them support the intrinsically. It’ll also improve testing on other platforms.

luhenry:
I've added the gcc and llvm builds, but I'm getting quite a few failures. I've explicitly disabled the configurations that don't work (no regression compared to `master`), but worth looking into some follow up PRs.

@blapie @GlassOfWhiskey let me know what you think

ericlove:
@blapie this looks good to me. Thanks to @luhenry for updating to the new intrinsics format!

luhenry:
I've enabled the LLVM builds for s390x and ppc64el. The issue was with cross-compilation of the builtin headers.

GlassOfWhiskey:
> I've added the gcc and llvm builds, but I'm getting quite a few failures. I've explicitly disabled the configurations that don't work (no regression compared to `master`), but worth looking into some follow up PRs.
> 
> @blapie @GlassOfWhiskey let me know what you think

I see that `riscv` builds with `gcc` are commented in the `.github/workflow` pipeline. Did they fail to compile for some reason?

luhenry:
> I see that riscv builds with gcc are commented in the .github/workflow pipeline. Did they fail to compile for some reason?

Yes, the gcc I'm getting from [riscv-gnu-toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain/releases/tag/2023.11.08) isn't recent enough, it needs to be GCC trunk and it's only GCC 13.2 at the moment. Once there is a new GCC release, we should be able to bump that version at [.github/workflows/build_and_test.yml#L145](https://github.com/shibatch/sleef/blob/3bcec26039771e8af513ebb184b80292345a633d/.github/workflows/build_and_test.yml#L145) and be able to test with GCC.

GlassOfWhiskey:
@luhenry it makes sense to me. For me the PR is ready to be merged. Maybe as a last bit, you could put in the `README.md` that SLEEF now supports `RISC-V V 1.0` with intrinsics `>=0.12`, just to make compatibility explicit.

GlassOfWhiskey:
Sorry Isee that README os quite minimal and doesn't say anything about compatibility. We csn put this in the changelog before next release

GlassOfWhiskey (approved):
LGTM

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
luhenry: The test is with Clang 17 given we are using LLVM version 17: https://github.com/shibatch/sleef/pull/477/files#diff-48c0ee97c53013d18d6bbae44648f7fab9af2e0bf5b0dc1ca761e18ec5c478f2R153

.github/workflows/build_and_test.yml
```diff
@@ -118,20 +138,38 @@ jobs:
...
-          sudo apt-get install -y -qq build-essential clang curl ninja-build libgmp-dev libmpfr-dev gcc${{ matrix.package || format('-{0}-linux-gnu', matrix.arch) }}
+          sudo apt-get install -y -qq build-essential curl ninja-build libgmp-dev libmpfr-dev gcc${{ matrix.gnupkg || format('-{0}-linux-gnu', matrix.arch) }}
+
+      - name: Install gcc
+        run: |
+          RISCV_GNU_TOOLCHAIN_TAG="2023.11.08"
+          curl -L https://github.com/riscv-collab/riscv-gnu-toolchain/releases/download/${RISCV_GNU_TOOLCHAIN_TAG}/riscv64-glibc-ubuntu-20.04-gcc-nightly-${RISCV_GNU_TOOLCHAIN_TAG}-nightly.tar.gz |
+            tar xzf - -C /opt
+          echo "PATH=/opt/riscv/bin:$PATH" >> $GITHUB_ENV
+        if: ${{ matrix.compiler == 'gcc' && matrix.arch == 'riscv64' }}
```
blapie, line 149: When comes the time to bump to gcc 14 in order to activate test for riscv64, there is a chance we use a more straightforward build of gcc (as in more widely used). For the sake of clarity would you mind removing this step, or replacing the content by 
```
sudo apt-get install -qq gcc${{ matrix.gnupkg || format('-{0}-linux-gnu', matrix.arch) }}
```
for all architectures so it matches the name of the step better and we don't apt install gcc at all for riscv64 (for now).

.github/workflows/build_and_test.yml
```diff
@@ -172,26 +219,81 @@ jobs:
...
+          # - arch: aarch64
+          #   compiler: llvm
+          #   qemu_cpu: "max,sve=on,sve256=on"
+          # - arch: aarch64
+          #   compiler: llvm
+          #   qemu_cpu: "max,sve=on,sve512=on"
           # Aarch32
           - arch: armhf
+            compiler: gcc
+            binfmt: arm
```
blapie, line 249: Would you be able to use the same layout as in `build-cross` ? As in have:
```
      matrix:
        arch: [aarch64, armhf, ppc64el, s390x, riscv64]
        compiler: [gcc, llvm]
```
and specifying `qemu_cpu` and `binfmt` in a similar way as `gnupkg`, while excluding some llvm and gcc variants?

luhenry: Defining multiple `include` with only the `qemu_cpu` being different will not trigger different builds unfortunately. The easiest way to trigger all the builds we want is to do is this way.

luhenry: I'm trying something out at https://github.com/rivosinc/sleef/compare/upstream-riscv...upstream-riscv-2, I'll merge into this PR if it's working.

src/libm/CMakeLists.txt
```diff
@@ -60,6 +60,13 @@ elseif(SLEEF_ARCH_S390X)
     PURECFMA_SCALAR
     DSP_SCALAR
   )
+elseif(SLEEF_ARCH_RISCV64)
+  set(SLEEF_HEADER_LIST
+    RVVM1
+    RVVM2
```
blapie, line 66: Sorry I only noticed that. Doid you not wish to provide *NOFMA versions (and consequently making cinz-variants available to users)? Did you have a reason for that?
AFAICT without this feature users won't be able to have access to a bitwise reproducible implementation across architectures that include both riscv and an architecture without FMA.
You could add it later but I don't think it is much more complicated than what you already have.

.github/workflows/build_and_test.yml
blapie: Sure! Let me know how this goes!

luhenry: It doesn't work: https://github.com/rivosinc/sleef/actions/runs/6867005626/job/18676581199

luhenry:
[aarch64](https://github.com/shibatch/sleef/actions/runs/6870688901/job/18686326827#step:9:1940), [ppc64el](https://github.com/shibatch/sleef/actions/runs/6870688901/job/18686328241#step:9:1934), and [s390x](https://github.com/shibatch/sleef/actions/runs/6870688901/job/18686328634#step:9:2941) fail with GCC 12 with the following error:
```
tanf denormal/nonnumber test : arg = -0, test = 0, correct = -0
```

src/libm/CMakeLists.txt
luhenry: > AFAICT without this feature users won't be able to have access to a bitwise reproducible implementation across architectures that include both riscv and an architecture without FMA.

I'm not sure I understand what you mean by "include both riscv and an architecture without FMA"? Do you mean if the user is running on a riscv machine with RVV, but that doesn't support FMA? And actually, what is FMA in that context?

blapie: This option (cinz) is here to provide reproducibility across architectures. Say a user is running on both an old architecture with no support for fused-multiply accumulate (FMA) and a newer FMA-enabled architecture say SVE or RVV, but wants bit reproducible results across those 2 architectures then it won't be possible unless there is a version of the helpers where the FMA is explicitly disabled ({RVV,SVE,...}*NOFMA). That is because RVV will use FMAs in places the old architecture does not (making old arch most likely less accurate, because 1 more rounding per FMA).

blapie: I recommend implementing it, because many users (including the Unity game-engine) end up using SLEEF because this feature is available. 
This obviously comes at a performance cost, that users are usually happy to pay.

blapie: Obviously penalty is only on NOFMA and cinz routines, no impact on other routines.

luhenry: There is no RVV without FMA, so on RISC-V only that would be quite artificial. However, in the context of comparing RISC-V to other older architecture, I understand the need for NOFMA. @ericlove you want to tackle that? I can take a look but it's going to take longer :)

luhenry: I just added it with https://github.com/shibatch/sleef/pull/477/commits/395a6f64bdedec5635afebe6cccee230c2290ce3. Let me know if it looks good, it's passing tests but not sure it's done correctly.

ericlove: Thanks @luhenry for adding the NOFMA variants and @blapie for pointing out their absence. I had not added them due to laziness and also not having any need myself for bitwise reproducibility across architectures, but they should of course be there for completeness, and the exercise of adding them seems to have uncovered an important error.

luhenry: WIP at https://github.com/rivosinc/sleef/commits/upstream-riscv-2

blapie: Fix looking good so far. The use of CONFIG (generally in SLEEF) isn't ideal, but at least it is consistent with other extensions.

blapie: I'm surprised this wasn't picked up in tests... Could be a combination of things...

luhenry: Fixed with https://github.com/shibatch/sleef/compare/ce3f65a210a04bdb5cec31cd2f384a4e0078987d%5E...e1b59d388395c36a5fa3324fd40973370c47fa78

ericlove: Is your surprise because you think the tests should have reported an error that was above of the specified accuracy tolerance? If so, that error might have been masked by another mistake I made, which was that I did not correctly implement the `vmla*` functions for the NOFMA case, and was emitting FMAs for them anyways.  I see @luhenry has corrected this in his recent commits.

luhenry:
@blapie it should have everything now. Let me know what you think, thanks!

blapie:
> @blapie it should have everything now. Let me know what you think, thanks!

Will have one last pass today, but it looks good to me!

.github/workflows/build_and_test.yml
```diff
@@ -142,22 +192,31 @@ jobs:
             EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DENFORCE_VSX=ON -DENFORCE_VSX3=ON"
           elif [[ ${{ matrix.arch }} = "s390x" ]]; then
             EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DENFORCE_VXE=ON"
-            # Disable VXE2 support, QEMU doesn't support it
+            # Disable VXE2 support, QEMU doesn't support some instructions generated by gcc or llvm
             EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DDISABLE_VXE2=ON"
```
blapie, line 196: Comment and option don't seem to match. What is intent here? Was the problem resolved?

src/libm/CMakeLists.txt
luhenry: Adding support for tester3 is a WIP at https://github.com/rivosinc/sleef/tree/upstream-riscv-2. I got the tests to build and run, but it's failing on a few tests:
```
1/4 Test  #7: tester3rvvm1 .....................***Failed  153.57 sec
sincos u10 7ff4e4fa460faee9b03c42ebb5918f5a
sincos u10 3fe37f4eb805505152f2b14a22a9f94e

sincos u35 5258989dd1f21e6550e4f95d2e7fe46b
sincos u35 95a7b7f48c71febf10ec6eff796dd391

sinf u35 390f777eb73eb2d76914ad4a1d4ff0ba
sinf u35 f8f804eae1d9443103e81fec96293477

cosf u35 03f61afc17485903a0dc002ce0326b5d
cosf u35 f2f3d1c9f090cde9c02439608dc7066e

tanf u35 8ed99e88f25785e580648e649c17fb06
tanf u35 68d42ad1fb412e6b8be3853461e61213


2/4 Test  #4: tester3rvvm1nofma ................***Failed  183.22 sec
sincos u10 5ece9dc5126b287182ea568f53f11382
sincos u10 7c164edcaa45988f6165b653fc76c495

sincos u35 fc3732f1bd033e30a69117f850a1462c
sincos u35 38fe7e261e184ed8dbf432ce6bedc5c4

sinf u35 f45b5ded20a1746474b2f3ddc62c13c1
sinf u35 833d845950b9cbb025629fe4c040f8f6

cosf u35 0fbf8ef44169b5ba1a9e97afd4544fce
cosf u35 74d7f871a6553cd0019087895e2052ad

tanf u35 c19c7d1e49f68be66181934ece099de8
tanf u35 bbb7c092d017e96d2454a38a20687735


3/4 Test #13: tester3rvvm2 .....................***Failed  282.74 sec
sincos u10 7ff4e4fa460faee9b03c42ebb5918f5a
sincos u10 3fe37f4eb805505152f2b14a22a9f94e

sincos u35 5258989dd1f21e6550e4f95d2e7fe46b
sincos u35 95a7b7f48c71febf10ec6eff796dd391

sinf u35 390f777eb73eb2d76914ad4a1d4ff0ba
sinf u35 f8f804eae1d9443103e81fec96293477

cosf u35 03f61afc17485903a0dc002ce0326b5d
cosf u35 f2f3d1c9f090cde9c02439608dc7066e

tanf u35 8ed99e88f25785e580648e649c17fb06
tanf u35 68d42ad1fb412e6b8be3853461e61213


4/4 Test #10: tester3rvvm2nofma ................***Failed  392.52 sec
sincos u10 5ece9dc5126b287182ea568f53f11382
sincos u10 7c164edcaa45988f6165b653fc76c495

sincos u35 fc3732f1bd033e30a69117f850a1462c
sincos u35 38fe7e261e184ed8dbf432ce6bedc5c4

sinf u35 f45b5ded20a1746474b2f3ddc62c13c1
sinf u35 833d845950b9cbb025629fe4c040f8f6

cosf u35 0fbf8ef44169b5ba1a9e97afd4544fce
cosf u35 74d7f871a6553cd0019087895e2052ad

tanf u35 c19c7d1e49f68be66181934ece099de8
tanf u35 bbb7c092d017e96d2454a38a20687735
```

I don't understand enough of SLEEF to figure out where that could come from. @blapie any pointer?

blapie: Is it failing on all archs? Can you point to the failing GHA, I cannot find it in your fork.

blapie: I would not blame the riscv implementation of nofma, since most routines are passing, right? Or are they not tested?

Thinking out loud. The common denominator here seems to be trig functions, that 
- all come in DETERMINISTIC and !DETERMINISTIC flavours
- all use rempi(f) that is special-cased for certain vector extensions (however sin,cos,tan and sincosf don't fail...) https://github.com/shibatch/sleef/blame/69cfa7130f3f2bda838633071b1c63aa7d56f940/src/libm/sleefsimddp.c#L298 

What happens if you also special case RISCV(NOFMA) there too? Maybe it has to do with how RISCV flushes to zero or something.

It would be good to see what input fails to confirm that this is the type of issue we are dealing with.

luhenry: That seems to be a riscv64 specific issue: https://github.com/rivosinc/sleef/actions/runs/6893048894. I will try to investigate tomorrow, but any additional pointers would be greatly appreciated.

blapie: Good news, should be easier to debug. I'll try to have a look too, but I'm not super familiar with risc-v intrinsics.
Looks like the scalar variant passes at least!
I would look at what is specific to trig functions that gets you a different result between scalar and vector (nofma).
Also are the helpers/mappings in tester3.c definitely right, I'd presume so if most tests are passing?

What command line do you use to fetch artifacts, what is the artifact ID?
That didn't work for me
```
gh api -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" rivosinc/sleef/actions/artifacts/test-riscv64-llvm-11
```
I would need to see the cmake test log with the details of the failures.

luhenry: I generally just click on the artifacts on the UI, and scp it to my VM if I need.

blapie: > All tests but 4 do pass, and it was the first thing I checked. Not sure if there are tests for these as well just to make sure I didn't screw it up.

Unfortunately, unit tests isn't really a thing in SLEEF...

> I generally just click on the artifacts on the UI, and scp it to my VM if I need.

Lol, sorry I forgot to scroll.

blapie: Will have to run with ctest --output-on-failure to get more from log.

luhenry: > Will have to run with ctest --output-on-failure to get more from log.

That's already the case on GHA. And if you want the tests' output from CI, you can download the `test-riscv64-llvm-11` artifact for example.

luhenry: Some values differ in the sincos function. Exhaustive values: https://gist.github.com/luhenry/8d102921ae20ca3a4d9c1270facb1e5c.

I haven't figure out yet why that is, but looking into it.

luhenry: Looking closely into the data, the differences are on -1 on the righest bit. For example,

luhenry: The `vrint_*` were indeed faulty and I've got something passing locally. Testing on CI and pushing on this PR.

luhenry:
@blapie all tests are passing (including tester3) with the latest changes.

src/arch/helperrvv.h
```diff
@@ -742,20 +743,10 @@ static INLINE vint vcast_vi_i(int32_t i) {
...
+  return SLEEF_RVV_DP_VFNCVT_X_F_VI(vd, __RISCV_VXRM_RNU, VECTLENDP);
 }
 static INLINE vdouble vrint_vd_vd(vdouble vd) {
-  // It is not currently possible to safely set frm for intrinsics,
-  // so emulate round-to-nearest behavior
-  vdouble half = SLEEF_RVV_DP_VCAST_VD_D(0.5, VECTLENDP);
-  half = __riscv_vfsgnj(half, vd, VECTLENDP);
-  vdouble res = __riscv_vfadd(vd, half, VECTLENDP);
-  return __riscv_vfwcvt_f(__riscv_vfncvt_rtz_x(res, VECTLENDP), VECTLENDP);
+  return SLEEF_RVV_DP_VFCVT_F_X_VD(SLEEF_RVV_DP_VFCVT_X_F_VD_RM(vd, __RISCV_VXRM_RNU, VECTLENDP), VECTLENDP);
```
ericlove, line 749: I believe these should use `__RISCV_FRM_RNE` since [these instructions](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#1317-single-width-floating-pointinteger-type-convert-instructions) are controlled by `frm`, not the fixed-point rounding mode, as reflected in [their intrinsics' API](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/main/auto-generated/intrinsic_funcs/04_vector_floating-point_intrinsics.adoc#vector-floating-point-move-intrinsics).  Same with all the similar ones above.

src/libm/CMakeLists.txt
ericlove: @luhenry yeah at the time I wrote the original port, there were no intrinsics with an `frm` parameter, so I made this apparently numerically-incorrect hack, but since these have been added now, I'm happy to see your use of them.  Thanks for fixing this!

src/arch/helperrvv.h
luhenry: I tried with passing `__RISCV_FRM_RNE` here and it didn't work. Given on the non-vector floating-point rounding mode, RNE is 0, and on the vector one, RNU is 0, I just passed "0" aka RNU. Do you know of documentation for what RNU means? I couldn't find anything...

ericlove: It's quite interesting that you observe different behavior in the two cases, since both of those enum values should map to zero, and do when I tested them: https://godbolt.org/z/YhPWMeMsE.  I would be interested to see the preprocessed output with both values, and to know which toolchain you are using.

I agree that the fixed-point rounding mode semantics match my original code, but these instructions are not fixed-point and are not controlled by the `vxrm` register.

luhenry: Actually, I just realized you pointed to "FRM" while I was using "VXRM", sorry. Let me try that on Monday as well.

ericlove: No worries. Thanks for looking into this, and have a great weekend!

CMakeLists.txt
```diff
@@ -89,8 +89,9 @@ set(COSTOVERRIDE_SVE 10)
 set(COSTOVERRIDE_SVENOFMA 10)
 set(COSTOVERRIDE_RVVM1 10)
 set(COSTOVERRIDE_RVVM1NOFMA 10)
-set(COSTOVERRIDE_RVVM2 10)
-set(COSTOVERRIDE_RVVM2NOFMA 10)
+set(COSTOVERRIDE_RVVM2 20)
+set(COSTOVERRIDE_RVVM2NOFMA 20
+)
```
blapie, line 94: nit: can we please close parenthesis on same line?

luhenry: Done.

luhenry:
@blapie should be fixed.

blapie:
LGTM - Will merge now!

blapie:
Done. Thanks for a brilliant collab! I hope we can add support for the rest of the features soon.

luhenry:
Yes, I already have some ideas on what I want to improve for RISC-V. Thanks again!

PR #489, opened 2023-12-11
Switch from old to new MD5 API in tester3.

### LINKED ISSUE #486: Deprecated usage of MD5 API
We currently get A LOT of warnings when compiling tester3 using recent version of OpenSSL.
That is because we are using the low level API, that's deprecated since OpenSSL 3.0.
We should switch to using the high level API.

blapie:
Removes warnings about deprecation of old API for versions of OpenSSL above 3.0.
Needed to amend openssl link line to avoid a crash in tester3 with old openssl.
Fixes #486.


blapie:
This patch also implements a workaround for this issue https://github.com/openssl/openssl/issues/13872

PR #491, opened 2023-12-12
Support parallel builds with make

joeramsay:
This patch series 'atomises' custom_commands in the cmake, such that there is no custom_command that contains multiple COMMANDs that have dependencies. This fixes the issue that `make` was not able to resolve these dependencies properly, and would sometimes create unusable headers if doing a parallel build.

Minimum CMake version is bumped to 3.18 in order to be able to use `cat` - this is a really common operation after atomising the commands.


joeramsay:
I think this PR probably subsumes #428, as all uses of `cat`/`type` are replaced with `cmake -E cat`, which should be cross-platform

blapie:
Merging this but we need to add test with GNU make in CI #494 .

PR #492, opened 2023-12-13
Fix build error with IntelLLVM compiler

iskunk:
I am building sleef using the IntelLLVM (oneAPI) 2024.0.0 compiler. This error comes up immediately:
```
$ make
[  0%] Building C object src/libm/CMakeFiles/mkrename_gnuabi.dir/mkrename_gnuabi.c.o
error: unknown argument '--extended_float_type'; did you mean '--extended_float_types'?
make[2]: *** [src/libm/CMakeFiles/mkrename_gnuabi.dir/build.make:76: src/libm/CMakeFiles/mkrename_gnuabi.dir/mkrename_gnuabi.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:838: src/libm/CMakeFiles/mkrename_gnuabi.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

If I add the "s" to the offending flag, I see this later in the build:
```
icx: warning: overriding '-ffp-model=fast=2' option with '-ffp-model=fast' [-Woverriding-t-option]
```
[This page](https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2024-0/fp-model-fp.html) in Intel's documentation does not indicate any numeric suboption to `-fp-model=fast`.

This PR adds a conditional to `Configure.cmake` that uses the modified values for the IntelLLVM variant. With this change, I get a clean build.

Note: Intel online community posts imply that the Intel Classic compiler may accept the `--extended_float_types` (plural) option in addition to the singular form, but I do not have a way of verifying this.

PR #497, opened 2024-01-02
fix intel compiler args.

xuhancn:
image
Intel compiler build failed.

image

Intel compiler only support "--extended_float_types" but not "--extended_float_type".

image
After fixed, it works.


xuhancn:
@shibatch Please review this PR.

blapie:
Hello,

Thanks for the suggestion!
This seems to be a duplicate of https://github.com/shibatch/sleef/pull/492.
We don't have access to Intel compilers, but this seems like a relatively generic change and seems acceptable.
I'll look into merging it this week!

xuhancn:
> Hello,
> 
> Thanks for the suggestion! This seems to be a duplicate of #492. We don't have access to Intel compilers, but this seems like a relatively generic change and seems acceptable. I'll look into merging it this week!

Thanks for reply.

blapie:
Merged #492 which should fix this issue! Cheers,

PR #503, opened 2024-01-17
Enable libsleefdft on riscv64

sh1boot:
Continuation of #496

Mostly just the in-fill of src/arch/helperrvv.h.


luhenry (changes requested):

src/arch/helperrvv.h
```diff
@@ -112,10 +122,23 @@ typedef vfloat32m4_t df2;
 typedef vint32m1_t vint2;
 typedef vint32m2_t fi_t;
 typedef vint32m4_t dfi_t;
+
 #define SLEEF_RVV_SP_LMUL 1
 #define SLEEF_RVV_DP_LMUL 1
+#if SLEEF_RVV_VLEN == 0
+#define VECTLENSP (__riscv_vsetvlmax_e32m1())
+#define VECTLENDP (__riscv_vsetvlmax_e64m1())
+static INLINE int vavailability_i(int name) {
```
luhenry, line 131: Define outside of this block, here is for macros and defines.

luhenry (changes requested):

src/arch/helperrvv.h
```diff
@@ -292,8 +341,11 @@ static INLINE vfloat figetd_vf_di(fi_t d) {
 static INLINE vint2 figeti_vi2_di(fi_t d) {
   return SLEEF_RVV_SP_VGET_VI2(d, 1);
 }
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wuninitialized"
 static INLINE fi_t fisetdi_fi_vf_vi2(vfloat d, vint2 i) {
   fi_t res;
+  //res = SLEEF_RVV_SP_VLMUL_EXT_VI2(SLEEF_RVV_SP_VREINTERPRET_VI2(d));
```
luhenry, line 348: Remove leftovers. Same for the `#pragma` right above and the other comments below.

src/arch/helperrvv.h
```diff
@@ -548,6 +603,17 @@ static INLINE vmask vand_vm_vo32_vm(vopmask x, vmask y) {
 static INLINE vmask vandnot_vm_vo32_vm(vopmask x, vmask y) {
   return __riscv_vmerge(y, 0, x, VECTLENSP);
 }
+#if defined(ENABLE_RVVM1) || defined(ENABLE_RVVM1NOFMA)
```
luhenry, line 606: If you can't use macros given the operation is different between RVVM1 and RVVM2, please put the #ifdef _inside_ the function to at least keep the signature in common.

luhenry (changes requested):

src/arch/helperrvv.h
sh1boot: This is where I started to fix the compilation warnings, so rather than remove them I doubled down and finished what I had started.  Yay no compiler warnings!

src/arch/helperrvv.h
```diff
@@ -72,30 +72,30 @@
...
+typedef vuint64m1_t rvv_dp_vmask;
+typedef vbool32_t rvv_sp_vopmask;
+typedef vbool64_t rvv_dp_vopmask;
 #elif defined(ENABLE_RVVM2) || defined(ENABLE_RVVM2NOFMA)
-typedef vuint64m4_t vmask;
-typedef vbool16_t vopmask;
+typedef vuint64m4_t rvv_sp_vmask;
+typedef vuint64m2_t rvv_dp_vmask;
+typedef vbool16_t rvv_sp_vopmask;
+typedef vbool32_t rvv_dp_vopmask;
```
sh1boot, line 84: I needed these extra types because I could not use the `ENABLE_RVV_SP` macro to disable some functions which used them internally, because they were used externally.

sh1boot:
Enabling inline headers will come next: https://github.com/rivosinc/sleef/pull/4

It refactors this code a little bit, but I assume it's better to draw a line under this before moving on.

luhenry (changes requested):

luhenry (approved):

luhenry:
@blapie could we please get a review on that, IMO it should be ready for RISC-V. Also it's passing on CI: https://github.com/rivosinc/sleef/actions/runs/7596179815

blapie:
Hi! Will have a look this week! Thanks for the contribution already!

src/dft/vectortype.h
```diff
@@ -37,6 +37,14 @@
...
 #endif
 
+#ifdef ENABLE_RVVM1
+#include "helperrvv.h"
+#endif
+
+#ifdef ENABLE_RVVM2
+#include "helperrvv.h"
+#endif
+
```
blapie, line 47: Please use
#if defined(ENABLE_RVVM1) || defined (ENABLE_RVVM2)

.github/workflows/build_and_test.yml
```diff
@@ -17,7 +17,7 @@ permissions:
 env:
   GCC_VERSION: "11"
   LLVM_VERSION: "17"
-  COMMON_CMAKE_FLAGS: |
+  COMMON_CMAKE_FLAGS: >
```
blapie, line 20: Just out of curiosity what does it change?

blapie (commented):
Looks sensible to me, but might need to have a second look because atm I'm mostly just trusting you on the technicalities.

.github/workflows/build_and_test.yml
luhenry: It's YAML syntax, it allows the following multi-line string to be single line. See https://yaml-multiline.info/

src/dft/vectortype.h
```diff
@@ -37,11 +37,7 @@
 #include "helpersve.h"
 #endif
 
-#ifdef ENABLE_RVVM1
-#include "helperrvv.h"
-#endif
-
-#ifdef ENABLE_RVVM2
+#if defined(ENABLE_RVVM1) || defined(ENABLE_RVVM1NOFMA) || defined(ENABLE_RVVM2) || defined(ENABLE_RVVM2NOFMA)
```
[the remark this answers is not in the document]
sh1boot: Oh, I see what you mean.  _Should_ it be that way?  Is DFT without FMA an inappropriate configuration?  Like, maybe it won't get past the unit tests or something?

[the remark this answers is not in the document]
blapie: I think it would be cleaner that way, as it is not enabled for any other vector extension (in this fail or in CMakeLists.txt). That is because *NOFMA extensions are here for reproducibility in libm routines, there is no such thing as reproducible/deterministic DFT in SLEEF, I believe.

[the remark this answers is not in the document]
sh1boot: OK.  So, by implication I should also remove the NOFMA variants from the libquad work that I'm doing right now, then?

sh1boot:
I'm not sure about some of the cast/truncate implementations, here.  vint is a different size from vmask (not sure why, is that ok?), and when casting between the two there are a mix of different policies for which bits to throw away.

Unfortunately if I change the implementation the tests still pass.

What's the expectation for functions that go between vint and vmask?

blapie:
> I'm not sure about some of the cast/truncate implementations, here. vint is a different size from vmask (not sure why, is that ok?), and when casting between the two there are a mix of different policies for which bits to throw away.

What do you mean by "there are a mix of different policies"? Do you mean cross vector extensions in SLEEF?
Not sure, if vint and vmask are meant to be of the same size or not, each vector extension could handle them differently. SVE for instance uses the same datatype for both, but it drops the most significant bits I believe, maybe you could use the same approach for truncating/casting?

> Unfortunately if I change the implementation the tests still pass.
> 
> What's the expectation for functions that go between vint and vmask?

I'm not aware of any particular expectation, as long as the choice is consistent, but my understanding of the dft part of the library is limited.

src/dft/vectortype.h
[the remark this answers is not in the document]
blapie: Yes, I don't see an occurrence of NOFMA in currently dispatched vector extension for quad, so not needed there either.

ericlove:
> I'm not sure about some of the cast/truncate implementations, here. vint is a different size from vmask (not sure why, is that ok?), and when casting between the two there are a mix of different policies for which bits to throw away.

Sorry for taking so long to join this conversation, but just thought it might be useful to chime in with my memory of why `vint` and `vmask` are the way they are in the RVV port (_warning_: this memory is reconstructed because it was long paged out...take it with a grain of salt).

The `vmask` type is constrained by the definition of [vcast_vm_i_i](https://github.com/shibatch/sleef/blob/df84fc7b31ab3982f4c705fe838a184414385d9f/src/arch/helperrvv.h#L527) to require 64-bit elements, even in single precision code in `sleefsimdsp.c`.  Unfortunately, this interacts badly with one of the design principles behind the RVV port, which is that all types should have the same VLMAX (`VECTLENSP` here), meaning that, for SP (assuming RVVM1 for example), `vmask` needs to have LMUL=2, while for DP, where the base type `vdouble` (`vfloat64m1_t`) already has 64-bit elements, it needs LMUL=1.

Meanwhile, `vint` follows a different logic.  It is used only in the DP codes, but for some reason is constrained to have 32-bit elements.  I no longer now why this is--either I ran into bugs in the original port when trying to widen it to 64 bits, or you have just found an opportunity to fix a major incoherence in the RVV design; you'll have to try it out to know which (and either way, we can certainly add the glaringly absent comment that should accompany it).  But given this is the case now, it means that to keep VLMAX constant at `VECTLENDP`, `vint` needs to have half the LMUL of other 64-bit types, including `vmask`.

Hopefully one of these constraints I described is unnecessary, and can be relaxed to simplify both the original RVV design, and your work on the DFT extensions.

sh1boot:
To clarify what I mean about 'different policies', here's the list of conversions to or from the vmask type that I found (supposing a configuration where m1 represents the minimum 128 bits):
```
vreinterpret_vf_vm: vncvt_x  256 -> 128
vreinterpret_vm_vf: vwcvtu_x  128 -> 256
vreinterpret_vd_vm: bitcast  128 -> 128
vreinterpret_vm_vd: bitcast  128 -> 128
vreinterpret_vm_vi64: bitcast  128 -> 128
vreinterpret_vi64_vm: bitcast  128 -> 128
vreinterpret_vm_vu64: bitcast  128 -> 128
vreinterpret_vu64_vm: bitcast  128 -> 128

vcast_vm_i64: splat  64 -> 128
vcast_vm_u64: splat  64 -> 128
vcast_vm_i_i: splat  (32+32) -> 128  or (32+32) -> 256 for single-precision
vcast_vm_vi: vwcvt_x  64 -> 128
vcastu_vm_vi: vwcvtu_x  64 -> 128
vcast_vi_vm: vncvt_x  128 ->64
vcastu_vi_vm: vnsrl 32  128 -> 64
```

Here I think `vncvt_x` will truncate on a per-lane basis (64-bit down to 32-bit for both single and double precision configurations), `vwcvt_x` will sign extend (and `vwcvtu_x` will zero-extend) on a per-lane basis (32-bit up to 64-bit), and `vnsrl 32` will truncate on a per-lane basis like `vncvt` but selecting the top 32 bits instead of the bottom 32 bits.  `splat` just repeats a 64-bit pattern until the destination is full.

Though that's in my local work tree after I've refactored a bit.  Comparing to the version in the current pull request, I see two implementations for `vcast_vi_vm()`, and the single-precision one is a truncate operation which is _definitely wrong_.  However, I can't think where that came from (evidently I added it but I don't see why I would because I don't think it's used?).

`vcast_vm_i_i` gives me pause, because it takes a pair of 64-bit inputs but still only creates a 64-bit pattern out of those before splatting.

It looks like `vcastu_vi_vm(vcastu_vm_vi(x))` would always return zero, which seems surprising as well.

But all the tests pass.

I'll try rationalising the size of `vmask` and `vint` for single-precision and see which options work.  If there's a more conventional arrangement then I guess I'll switch to that instead.

sh1boot:
> The vmask type is constrained by the definition of [vcast_vm_i_i](https://github.com/shibatch/sleef/blob/df84fc7b31ab3982f4c705fe838a184414385d9f/src/arch/helperrvv.h#L527) to require 64-bit elements, even in single precision code in sleefsimdsp.c.

The way it's used in sleefsimdsp.c it looks like it's just creating pairs of complementary floats, so it's not unreasonable to cast it back to a 32-bit type having the same VECTLENSP as float32 operations -- then it works out to be the same bit width for DP and SP cases.  It just needs a temporarily 64-bit splat while the pair of floats are unrolled.

Working that change through piece by piece it looks like all (or just most?) of the compatibility workarounds between SP and DP disappear.  So I guess I'll do that.

sh1boot:
Looks like the first step in rebasing automatically closed this PR?  Anyway, I've rebased it, now, and fixed or understood everything that I was worried about.

.github/workflows/build_and_test.yml
```diff
@@ -227,8 +227,6 @@ jobs:
             EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DSLEEF_ENFORCE_RVVM1=ON -DSLEEF_ENFORCE_RVVM2=ON"
             # Disable inline headers, they just don't compile on riscv64
             EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DSLEEF_BUILD_INLINE_HEADERS=OFF"
-            # Disable dft, it fails with linker error to `cexp`
-            EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DSLEEF_BUILD_DFT=OFF"
             # Disable quad, it's missing the `Sleef_quad` function
             EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DSLEEF_BUILD_QUAD=OFF"
```
luhenry, line 231: If you're adding support for quad at https://github.com/shibatch/sleef/pull/503/files#diff-cced2a666d218fe891973b1e7d9ef8b64c4467b511ab815073c4fc670bc91a31R1152, you will also need to remove this line to make sure it's tested on CI.

sh1boot: I moved the intrinsics implementations for quad into this patch by accident, but then decided I might as well leave them there (at the risk of making this an ever-growing PR) so that we don't have two rounds of reviewing the intrinsics implementation themselves.

I was going to say, I have a bigger [patch for quad](https://github.com/rivosinc/sleef/compare/dev/shosie/fix-inline-headers...rivosinc:sleef:dev/shosie/fix-rvv-quad) but looking at it now, I might just as well cherry-pick that into this patch, because it doesn't look like there's much left to object to.

The only problem is that the order I did things, those patches were written _after_ [inline-header support](https://github.com/rivosinc/sleef/compare/dev/shosie/fix-rvv-dft...rivosinc:sleef:dev/shosie/fix-inline-headers).  My approach there is more questionable.  Mostly because of the complexity of handling the incompatibility between SP and DP implementations of some functions.

The size of that PR has also become much smaller after rework, here, but it still has the same design issues (eg., emitting conditional preprocessor logic in the generated macroonly*.h files).

Let's see if I can't reorder the patches and fold the rest of quad into this....

sh1boot: Looks like I can indeed!

sh1boot:
I've carved out some of this patch into #520 and #521.  I'm not sure how github works, here, but I assume that this patch now depends on those PRs going in first, and when they do this diff will shrink?

blapie:
@sh1boot This one need a rebase now.

luhenry:
@blapie I've rebased it and updated this PR. The diff between before and after the rebase is the same: https://github.com/shibatch/sleef/compare/82411c1f1883a528471db11c7439b56857038de8..b1c49df4ebe7163f6ab6c553586d0840a46f627e.

Issue #485, opened 2023-11-20
Some tests fail on AArch64 with LLVM 17


blapie:
PR https://github.com/shibatch/sleef/pull/477 pointed out some test failures with qemu.
Tests are currently disabled, thus limiting support for aarch64 to gcc.
Need to fix this!

blapie:
Not able to reproduce bug. Tests seem to pass on GHA and locally. 
@joanaxcruz re-enabled tests for AArch64 with llvm on Linux in #506.

PR #506, opened 2024-01-25
Enable aarch64 llvm-17 cross compilation tests

joanaxcruz:
PR #477 pointed out some test failures with qemu.
However, following the discussion on this PR cannot see exactly what these failures were (none of the GHA past builds show runs for these settings). Locally, cross compiling aarch64 on x86 host with llvm-17 and using user-mode qemu emulation environment do not
show failures.
Therefore we try to enable these tests in GHA and check for failures.
Note: This issue is being tracked in #485


joanaxcruz:
No failures reported. Changes should stay to enable testing aarch64 when cross compiled with llvm17 on x86 host. Ready to merge and close PR.

PR #508, opened 2024-01-25
Add qemu command prefix to testervecabi tests

joanaxcruz:
This test suite covers the usage of scalar symbols with auto-vectorisation, to test the GNU ABI
symbols generated by SLEEF.
However, some failures are recorded when emulating these tests using user mode QEMU, for a
cross-compile build for AArch64 and s390x.
This is because this test suite does not get called with the EMULATOR command necessary.
This fix adds the emulator to the test command
which makes the tests pass when cross-compiling
and using user mode QEMU emulation.

### DIFF
--- a/src/libm-tester/CMakeLists.txt
+++ b/src/libm-tester/CMakeLists.txt
@@ -462,7 +462,7 @@ if(ENABLE_GNUABI AND COMPILER_SUPPORTS_OMP_SIMD AND NOT SLEEF_TARGET_PROCESSOR M
   target_compile_options(testervecabi PRIVATE ${OpenMP_C_FLAGS})
   target_link_libraries(testervecabi ${TARGET_LIBSLEEF} ${OpenMP_C_FLAGS})
   set_target_properties(testervecabi PROPERTIES C_STANDARD 99)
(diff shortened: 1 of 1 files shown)


joanaxcruz:
A bit more on this issue. 

Looking at the logs from running 
`ctest --output-on-failure --test-dir build-aarch64-gcc`
on a aarch64 cross build on a x86 host where we pass -DEMULATOR=qemu-aarch64:

Before these changes, testervecabit suite would be invoked like:
`Test command: "$(PWD)/sleef/build-aarch64-gcc/bin/testercevabi"`

For some reason when these tests were being run in a docker environment, the following errors would show:
image

After these changes, the invocation of testervecabi becomes:
`Test command: /usr/bin/qemu-aarch64 "$(PWD)/sleef/build-aarch64-gcc/bin/testercevabi"`

We stop observing the error above.

blapie:
Maybe worth mentioning that these failures were not observed on GHA since QEMU is invoked on a system level, and no EMULATOR is passed to cmake.
Maybe something to change in the future, so emulation is enabled as expected by build system, i.e. via a prescribed EMULATOR.

PR #510, opened 2024-02-07
add static build config.

xuhancn:
1. Add "SLEEF_BUILD_SHARED_LIBS" to instead of CMake reserved variable "BUILD_SHARED_LIBS".
2. Add clear library type to add_library, it will remove the global control from "BUILD_SHARED_LIBS", Link: https://cmake.org/cmake/help/latest/guide/tutorial/Selecting%20Static%20or%20Shared%20Libraries.html 
3. remove install to dummy dir, it will make git dirty.

Why we avoid to use `BUILD_SHARED_LIBS` and `add_library` without `STATIC/SHARED` parameter. Please check the offical document `Note` part.
image

Additional, I write a test project to test this PR: https://github.com/xuhancn/sleef_cmake_example
Turn on/off this line to control the build type.
https://github.com/xuhancn/sleef_cmake_example/blob/main/CMakeLists.txt#L14 

Static：
image
Shared:
image


xuhancn:
@blapie Please comment and review.

xuhancn:
It seems that, we need this PR to fix pytorch sleef build issue. https://github.com/pytorch/pytorch/pull/118980

blapie:
Hello! I think I understand the issue/suggested solution, although it did not sound extremely clear. 

I don't really understand how this works as is. While the change to CMakeLists.txt seems to make sense, in Configure.cmake and src/common/CMakeLists.txt occurrences of BUILD_SHARED_LBS are replaced by SLEEF_STATIC_LIBS, which I believe is the opposite of what you want. Btw SLEEF_BUILD_STATIC_LIBS is not defined as a user-facing option, which doesn't help understanding the solution.

I'm not very keen on that change anyway, even if it was fixed. The use case is too particular to justify making the build system more complex than it already is.

If pytorch needs to setup its submodules differently locally then the solution implementing Save/Restore (https://github.com/pytorch/pytorch/blob/cee16353db384dd33a4804f9f26829cacfcfbab7/aten/src/ATen/CMakeLists.txt#L424-L440) was probably the way to go, or something along those lines.

I don't believe it is a relevant feature to add to SLEEF, but happy to be convinced otherwise.

xuhancn:
Also, you can check the another static lib mimalloc: https://github.com/pytorch/pytorch/blob/main/CMakeLists.txt#L1060-L1068
> Setup mimalloc dedicate options.
> Add mimalloc as submodule.
> Use setup include and link as depends.

Acturlly, save/restore is a workaround for sleef exist "BUILD_SHARED_LIBS" issue. Please check this comment: https://github.com/pytorch/pytorch/blob/cee16353db384dd33a4804f9f26829cacfcfbab7/aten/src/ATen/CMakeLists.txt#L483

xuhancn:
Finally, It is continued to rename "BUILD_SCALAR_LIB " to "SLEEF_BUILD_SCALAR_LIB ", please my last PR: https://github.com/shibatch/sleef/pull/509#issuecomment-1929181290
Also, this PR make a rule to enable "SLEEF_BUILD_SCALAR_LIB" to control sleef static/shared library.

xuhancn:
@blapie Please review it again.

xuhancn:
@blapie could you please priority to review this PR? https://github.com/pytorch/pytorch/pull/118980 this pytorch PR id depends on it.

blapie:
Hello @xuhancn! I'm looking into now. I would prefer a slightly different approach though but I will look at how feasible that is.

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 
 # Options
-
+option(SLEEF_BUILD_SHARED_LIBS "Build sleef as static library" ON)
```
blapie, line 3: Can you try this change instead and switch back to old name everywhere? I believe this can work according to the documentation (https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html)
```
-option(SLEEF_BUILD_SHARED_LIBS "Build sleef as static library" ON)
+# Declare BUILD_SHARED_LIBS option to make it accessible to top-level projects.
+option(BUILD_SHARED_LIBS "Build using shared libraries (CMake reserved)" ON)
+
```

blapie:
Right, let me know if I'm missing something but I believe we don't have to rename the variable to get the behaviour you are expecting.

Could you try defining the BUILD_SHARED_LIBS option in SLEEF's CMakeLists.txt see comment in review?
According to the documentation of CMake it should work, when top-level project switches the option.

https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html
```
Note that if bringing external dependencies directly into the build, such as with [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html#module:FetchContent) or a direct call to [add_subdirectory()](https://cmake.org/cmake/help/latest/command/add_subdirectory.html#command:add_subdirectory), and one of those dependencies has such a call to [option(BUILD_SHARED_LIBS ...)](https://cmake.org/cmake/help/latest/command/option.html#command:option), the top level project must also call [option(BUILD_SHARED_LIBS ...)](https://cmake.org/cmake/help/latest/command/option.html#command:option) before bringing in its dependencies. Failure to do so can lead to different behavior between the first and subsequent CMake runs.
```

https://discourse.cmake.org/t/build-shared-libs/4200
```
It’s fine to declare it as an option if you want to change the default. As long as the minimum required CMake version is 3.13, a parent project can override it with a normal variable.
```

But it might need to be done the right way. In order to fix the parallel build I think you need to put the save/restore stages in a function (since it will have a scope), see
https://stackoverflow.com/a/31140797

CMakeLists.txt
xuhancn: I have tried this method, It not constant work, the reason as above: https://github.com/shibatch/sleef/pull/510#issuecomment-1934336010

xuhancn: pytorch default want setup “BUILD_SHARED_LIBS” as shared, but it want to use sleef as static library. Current pytorch is used a save/restore this variable to WA it. 
Save: https://github.com/pytorch/pytorch/blob/cee16353db384dd33a4804f9f26829cacfcfbab7/aten/src/ATen/CMakeLists.txt#L424-L440
Restore: https://github.com/pytorch/pytorch/blob/cee16353db384dd33a4804f9f26829cacfcfbab7/aten/src/ATen/CMakeLists.txt#L480-L485
And the original author think we should use SLEEF_ to pass the sleef build config also: https://github.com/pytorch/pytorch/blob/cee16353db384dd33a4804f9f26829cacfcfbab7/aten/src/ATen/CMakeLists.txt#L483

xuhancn:
> Right, let me know if I'm missing something but I believe we don't have to rename the variable to get the behaviour you are expecting.
> 
> Could you try defining the BUILD_SHARED_LIBS option in SLEEF's CMakeLists.txt see comment in review? According to the documentation of CMake it should work, when top-level project switches the option.
> 
> https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html
> 
> ```
> Note that if bringing external dependencies directly into the build, such as with [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html#module:FetchContent) or a direct call to [add_subdirectory()](https://cmake.org/cmake/help/latest/command/add_subdirectory.html#command:add_subdirectory), and one of those dependencies has such a call to [option(BUILD_SHARED_LIBS ...)](https://cmake.org/cmake/help/latest/command/option.html#command:option), the top level project must also call [option(BUILD_SHARED_LIBS ...)](https://cmake.org/cmake/help/latest/command/option.html#command:option) before bringing in its dependencies. Failure to do so can lead to different behavior between the first and subsequent CMake runs.
> ```
> 
> https://discourse.cmake.org/t/build-shared-libs/4200
> 
> ```
> It’s fine to declare it as an option if you want to change the default. As long as the minimum required CMake version is 3.13, a parent project can override it with a normal variable.
> ```
> 
> But it might need to be done the right way. In order to fix the parallel build I think you need to put the save/restore stages in a function (since it will have a scope), see https://stackoverflow.com/a/31140797

The original pytorch is use this method, it works on Linux but not works on Windows. If It works, why I do a lot of work to fix it?

xuhancn:
> First, what I was getting at is not simply moving back to using BUILD_SHARED_LIBS but also making sure save/add_subdirectory/restore occurs inside a function (or at least a scope) so it builds safely in parallel. Have you tried that? This should be applied everywhere save/restore is used (not just for SLEEF).
> 
> > I can't explain the detail why it doesn't work. But I tried it and it not work. It is OS related.
> 
> Second, if the above approach does not work, maybe it would help to dive a little bit more into why this actually fails. We are not excluding that something else needs to be fixed in SLEEF to make pytorch build on Windows. Introducing SLEEF_BUILD_SHARED_LIBS is just not an appropriate change from our point of view. This blog is a good reference for this type of issues and it mentions the function-trick https://alexreinking.com/blog/building-a-dual-shared-and-static-library-with-cmake.html
> 
> If PyTorch's use of BUILD_SHARED_LIBS diverges from its definition (by building both static and shared components) then it needs to be handled properly in PyTorch and not in the components. Therefore, I would not refer to the save/restore trick as a "work around", or more precisely not refer to this as a SLEEF issue.

```
set(sources ...)
add_library(SomeLib_static STATIC ${sources})
add_library(SomeLib_shared SHARED ${sources})
```
From the example, gen sleef_static and sleef(shared) at the same time? do you think it is acceptable solution.

blapie:
Not sure I understand. I don't think this will solve the issue, but feel fee to try.

I was picturing defining a function encapsulating everything at this link https://github.com/pytorch/pytorch/blob/cee16353db384dd33a4804f9f26829cacfcfbab7/aten/src/ATen/CMakeLists.txt#L442-L478. Then just calling this function.

```
function(add_sleef)
 // basically all the content at link above
endfunction()

add_sleef()
```

The benefit of the function is that you don't even have to save and restore BUILD_SHARED_LIBS and BUILD_TESTS, the scope of the function itself protects from overriding  BUILD_SHARED_LIBS at the global scope.

xuhancn:
> Not sure I understand. I don't think this will solve the issue, but feel fee to try.
> 
> I was picturing defining a function encapsulating everything at this link https://github.com/pytorch/pytorch/blob/cee16353db384dd33a4804f9f26829cacfcfbab7/aten/src/ATen/CMakeLists.txt#L442-L478. Then just calling this function.
> 
> ```
> function(add_sleef)
>  // basically all the content at link above
> endfunction()
> 
> add_sleef()
> ```
> 
> The benefit of the function is that you don't even have to save and restore BUILD_SHARED_LIBS and BUILD_TESTS, the scope of the function itself protects from overriding BUILD_SHARED_LIBS at the global scope.

1. I means that If we can consider to build shared and static lib at the same time, like mimalloc: https://github.com/microsoft/mimalloc/blob/master/CMakeLists.txt#L22-L23
shared library name: sleef.
static library name: sleef_static.

2. I don't think cmake function can works. `BUILD_SHARED_LIBS` is global scope.
image
FROM: https://cmake.org/cmake/help/git-stage/variable/BUILD_SHARED_LIBS.html

Please read the offical guide. The offical guide said that: 
`do so in the top level CMakeLists.txt file, before any add_library() calls.` and `Note that if bringing external dependencies directly into the build, such as with FetchContent or a direct call to add_subdirectory(), and one of those dependencies has such a call to option(BUILD_SHARED_LIBS ...), the top level project must also call option(BUILD_SHARED_LIBS ...)before bringing in its dependencies. Failure to do so can lead to different behavior between the first and subsequent CMake runs.`

Use the `BUILD_SHARED_LIBS` will bring in a lot of limitation.

xuhancn:
Hi @blapie 
Acturaly, sleef library as a math library, it is usual been integated into another projects as a compontemt. And in git management system, it is usual as a git submodule.
From the CMake offical document of `BUILD_SHARED_LIBS`:
1. `do so in the top level CMakeLists.txt file, before any add_library() calls.` 
2. `Note that if bringing external dependencies directly into the build, such as with FetchContent or a direct call to add_subdirectory(), and one of those dependencies has such a call to option(BUILD_SHARED_LIBS ...), the top level project must also call option(BUILD_SHARED_LIBS ...)before bringing in its dependencies. Failure to do so can lead to different behavior between the first and subsequent CMake runs.`

In git-cmake build system, we usual build dependency via `add_subdirectory()`, As a submodule, sleef request its parent project also define `BUILD_SHARED_LIBS` ? whether it too much limitation?
pytorch is not and can't set `BUILD_SHARED_LIBS` due to it is a CMake reserved variable. And the doc said: `Failure to do so can lead to different behavior between the first and subsequent CMake runs.` is similar to my conclusion: https://github.com/shibatch/sleef/pull/510#issuecomment-1934336010

We maintains a open source project should consider easy to developer use. Use `BUILD_SHARED_LIBS` control sleef will bring in too much limitation, and impact to parent project and other sub-modules.
My PR is good to solve the issue:
1. "SLEEF_" prefix of "SLEEF_BUILD_SHARED_LIBS" shows it only control and impact on sleef project.
2. Not use CMake global varibal and not impact to other modules.
3. No limitation and easy for developer use.

blapie:
We have no issue with people using SLEEF as a submodule, if anything that is probably the preferred approach and we provide examples for that.

> As a submodule, sleef request its parent project also define BUILD_SHARED_LIBS ? whether it too much limitation?

No, you don't have to define it, but if you need to switch between static and shared libs then that is the flag you should use. 

Sorry I am going to repeat myself but your PR "solves" the problem from your point of view (or rather hides it) by introducing flaws in our codebase. We cannot accept a change like that simply because it will give unexpected results to SLEEF users (who also use it as a submodule).

Instead I suggested a simple alternative, that you don't want to try, which is encapsulating the options and `add_subdirectory(sleef)` in a function. I did it for you and it works, as far as your simple example goes, it does not modify BUILD_SHARED_LIBS in the global scope : https://github.com/xuhancn/sleef_cmake_example/pull/1

I don't see what's wrong with this approach
1. it is actually safer in terms of parallel build and global scope for pytorch
2. you don't need me to approve a SLEEF PR to fix your problem

xuhancn:
@blapie some days passed, any update?

blapie:
> [blapie_example_output.txt](https://github.com/shibatch/sleef/files/14483892/blapie_example_output.txt)
I attached the all log to you for analysis.

Btw I do not see the messages I added to the CMakeLists.txt in your log. It should show this, while `BUILD_SHARED_LIBS=OFF` is set within the function.

```
Set global BUILD_SHARED_LIBS in parent project
BUILD_SHARED_LIBS=ON
[...]
BUILD_SHARED_LIBS=ON
[...]
```

Are these runs your are showing 100% including my changes?

Based on the first couple of command lines I don't think it is including my PR, there should be at least a `gh pr checkout 1`.

blapie:
Ok, right in this case even the config log seems wrong
```
-- Building shared libs : ON
```
although the top-level CMakeLists.txt has
```
set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build sleef as static library" FORCE)
```

Can you please provide a similar log with the master branch of SLEEF?

xuhancn:
> Hi! Sorry, I did not find time to test before. But I just did it on MSYS2 (Unix-like Windows environment) and as expected I get static libs when setting `BUILD_SHARED_LIBS=OFF` in the function, and I get shared libs otherwise. This is a Windows on Arm environment and I'm using Clang (not cl) so not exactly your case, but the difference should be quite small.
> 
> It would be good to go to the bottom of that, as it could simply be that a minor change in SLEEF is needed to work on Windows on x86 with cl.
> 
> Please note that SLEEF support for MSVC/CL is limited for now, in particular reproducing and testing is not available on the repo.

This PR is small change to enable msvc cl, which is Microsoft offical compiler.
1. We only add a `SLEEF_` prefix to `BUILD_SHARED_LIBS` and it is align to https://github.com/shibatch/sleef/pull/509
2. Use `SLEEF_BUILD_SHARED_LIBS` to control build shared/static library straightforward.

MSVC should be the fisrt tier compiler on Windows, we need enable it ASAP.

xuhancn:
> Ok, right in this case even the config log seems wrong
> 
> ```
> -- Building shared libs : ON
> ```
> 
> although the top-level CMakeLists.txt has
> 
> ```
> set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build sleef as static library" FORCE)
> ```
> 
> Can you please provide a similar log with the master branch of SLEEF?

Checkout sleef submodule to master, and collect log again?

xuhancn:
> > Ok, right in this case even the config log seems wrong
> > ```
> > -- Building shared libs : ON
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > although the top-level CMakeLists.txt has
> > ```
> > set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build sleef as static library" FORCE)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Can you please provide a similar log with the master branch of SLEEF?
> 
> Checkout sleef submodule to master, and collect log again?

[blapie_debug_log_sleef_master.txt](https://github.com/shibatch/sleef/files/14605029/blapie_debug_log_sleef_master.txt)
Log is here.

xuhancn:
> Ok, so we have the expected behaviour right? We build the static library while BUILD_SHARED_LIBS=ON is unchanged at the global scope.
> 
> ```
> sleefscalar.vcxproj -> D:\xuhan\build_pytorch\sleef_cmake_example\build_3\third_party\sleef\lib\sleefscalar.lib
> ```

It works, but pytorch is use default value and it's value is OFF. You mean let pytorch use `function(add_sleef)` to integrate sleef? 
I don't known if it works, I think is not a simple & straightforward method.

xuhancn:
> > Ok, so we have the expected behaviour right? We build the static library while BUILD_SHARED_LIBS=ON is unchanged at the global scope.
> > ```
> > sleefscalar.vcxproj -> D:\xuhan\build_pytorch\sleef_cmake_example\build_3\third_party\sleef\lib\sleefscalar.lib
> > ```
> 
> It works, but pytorch is use default value and it's value is OFF. You mean let pytorch use `function(add_sleef)` to integrate sleef? I don't known if it works, I think is not a simple & straightforward method.

I tried to use cmake function integate sleef into pytorch and it failed at binary size.
image
PR: https://github.com/xuhancn/pytorch/pull/8
Log:
[pytorch_add_sleef_function.txt](https://github.com/shibatch/sleef/files/14612250/pytorch_add_sleef_function.txt)

blapie:
Sounds very strange. I don't see the connection between the diff and the error message.

blapie:
Should https://github.com/xuhancn/pytorch/pull/8 be rebased first?

xuhancn:
> Should [xuhancn/pytorch#8](https://github.com/xuhancn/pytorch/pull/8) be rebased first?

No, it is passed if use SLEEF_BUILD_SHARED_LIB.

xuhancn:
@jgong5 I have update code to compatible to `BUILD_SHARED_LIB`, and also work for `SLEEF_BUILD_SHARED_LIB`.
@blapie please review and merge ASAP.

blapie:
I'm looking into your last commit and will leave comments in patch.

Meanwhile let me go back to the initial problem for a second.

I just cannot believe that CMake created a variable that is not usable when doing such trivial thing as adding components as submodules. In fact what CMake says is (sorry for quoting again but this is pretty relevant to our position)
```
Note that if bringing external dependencies directly into the build, such as with [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html#module:FetchContent) or a direct call to [add_subdirectory()](https://cmake.org/cmake/help/latest/command/add_subdirectory.html#command:add_subdirectory), and one of those dependencies has such a call to [option(BUILD_SHARED_LIBS ...)](https://cmake.org/cmake/help/latest/command/option.html#command:option), the top level project must also call [option(BUILD_SHARED_LIBS ...)](https://cmake.org/cmake/help/latest/command/option.html#command:option) before bringing in its dependencies. Failure to do so can lead to different behavior between the first and subsequent CMake runs.
```
The "Failure to do so" refers to the need to add `option(BUILD_SHARED_LIBS ...)` before `add_subdirectory(sleef)`. I don't think at any point we tried to follow this indication and add `option(BUILD_SHARED_LIBS ...)` rather than `set(BUILD_SHARED_LIBS ...)` before `add_subdirectory(sleef)`.

Could it be a simple solution to the issue? Could you please try that?

xuhancn:
> I'm looking into your last commit and will leave comments in patch.
> 
> Meanwhile let me go back to the initial problem for a second.
> 
> I just cannot believe that CMake created a variable that is not usable when doing such trivial thing as adding components as submodules. In fact what CMake says is (sorry for quoting again but this is pretty relevant to our position)
> 
> ```
> Note that if bringing external dependencies directly into the build, such as with [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html#module:FetchContent) or a direct call to [add_subdirectory()](https://cmake.org/cmake/help/latest/command/add_subdirectory.html#command:add_subdirectory), and one of those dependencies has such a call to [option(BUILD_SHARED_LIBS ...)](https://cmake.org/cmake/help/latest/command/option.html#command:option), the top level project must also call [option(BUILD_SHARED_LIBS ...)](https://cmake.org/cmake/help/latest/command/option.html#command:option) before bringing in its dependencies. Failure to do so can lead to different behavior between the first and subsequent CMake runs.
> ```
> 
> The "Failure to do so" refers to the need to add `option(BUILD_SHARED_LIBS ...)` before `add_subdirectory(sleef)`. I don't think at any point we tried to follow this indication and add `option(BUILD_SHARED_LIBS ...)` rather than `set(BUILD_SHARED_LIBS ...)` before `add_subdirectory(sleef)`.
> 
> Could it be a simple solution to the issue? Could you please try that?

My solution is merge this PR.
You can setup a Windows MSVC environment and try it by yourself. Thanks.

blapie:
@alexreinking Hello! Sorry to bother you, but we stumbled upon your blog (https://alexreinking.com/blog/building-a-dual-shared-and-static-library-with-cmake.html) and wondered if maybe you could shine some light on this situation?
There seems to be issues controlling BUILD_SHARED_LIBS at the component boundary in pytorch (https://github.com/pytorch/pytorch/blob/cee16353db384dd33a4804f9f26829cacfcfbab7/aten/src/ATen/CMakeLists.txt#L423-L485).
We would like to avoid merging a change like currently suggested in the PR, as we think it will confuse users since it will result in unexpected behaviour.
What do you recommend?

CMakeLists.txt
```diff
@@ -29,6 +28,13 @@ option(SLEEF_DISABLE_SSL "Disable testing with the SSL library" OFF)
 option(SLEEF_ENABLE_CUDA "Enable CUDA" OFF)
 option(SLEEF_ENABLE_CXX "Enable C++" OFF)
 
+set(_INTERNAL_SLEEF_BUILD_SHARED_LIBS)
+if(NOT DEFINED ${SLEEF_BUILD_SHARED_LIBS})
+  set(_INTERNAL_SLEEF_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
+else()
+  set(_INTERNAL_SLEEF_BUILD_SHARED_LIBS ${SLEEF_BUILD_SHARED_LIBS})
+endif()
+
```
blapie, line 37: Can this be done with options instead? For instance, 
```
option(BUILD_SHARED_LIBS "Build sleef as shared library" ON)
option(SLEEF_BUILD_SHARED_LIBS "Build sleef as shared library" ${BUILD_SHARED_LIBS})
```

alexreinking: @blapie -- These are not equivalent because this defines `SLEEF_BUILD_SHARED_LIBS` as a cache variable when no variable by that name (normal or cache) exists. So then changing `BUILD_SHARED_LIBS` only works as expected on the first configuration.

My blog post suggests the following:

```suggestion
if (DEFINED SLEEF_BUILD_SHARED_LIBS)
  set(BUILD_SHARED_LIBS ${SLEEF_BUILD_SHARED_LIBS})
endif ()
```

Then you can just look at the truthiness of `BUILD_SHARED_LIBS` down the line. If you care about documenting this variable in the CMake CLI/GUI, then this approach works:

```suggestion
set(SLEEF_BUILD_SHARED_LIBS "undefined"
    CACHE BOOL "Override BUILD_SHARED_LIBS while configuring SLEEF")
if (NOT SLEEF_BUILD_SHARED_LIBS STREQUAL "undefined")
  set(BUILD_SHARED_LIBS ${SLEEF_BUILD_SHARED_LIBS})
endif ()

File: CMakeLists.txt [hunk 3]
```diff
@@ -157,9 +163,9 @@ separate build directory. Note: Please remove autogenerated file \
 `CMakeCache.txt` and directory `CMakeFiles` in the current directory.")
 endif()
 
-if(SLEEF_ENABLE_LTO AND BUILD_SHARED_LIBS)
-  message(FATAL_ERROR "SLEEF_ENABLE_LTO and BUILD_SHARED_LIBS cannot be specified at the same time")
-endif(SLEEF_ENABLE_LTO AND BUILD_SHARED_LIBS)
+if(SLEEF_ENABLE_LTO AND _INTERNAL_SLEEF_BUILD_SHARED_LIBS)
+  message(FATAL_ERROR "SLEEF_ENABLE_LTO and _INTERNAL_SLEEF_BUILD_SHARED_LIBS cannot be specified at the same time")
+endif(SLEEF_ENABLE_LTO AND _INTERNAL_SLEEF_BUILD_SHARED_LIBS)
```
alexreinking, line 168: With my other suggestions:

```suggestion
if (SLEEF_ENABLE_LTO AND BUILD_SHARED_LIBS)
  message(FATAL_ERROR "SLEEF_ENABLE_LTO and (SLEEF_)BUILD_SHARED_LIBS cannot be enabled at the same time")
endif ()
```
Note that duplicating the condition in the `endif` hasn't been a requirement since at _least_ 2.8.x but it might be as far back as 2.4.x. Definitely not needed in 3.18, which is your minimum.

CMakeLists.txt
```diff
@@ -305,7 +311,7 @@ if(SLEEF_SHOW_CONFIG)
     message("   Native build dir: ${NATIVE_BUILD_DIR}")
   endif(CMAKE_CROSSCOMPILING)
   message(STATUS "Using option `${SLEEF_C_FLAGS}` to compile libsleef")
-  message(STATUS "Building shared libs : " ${BUILD_SHARED_LIBS})
+  message(STATUS "Building shared libs : " ${_INTERNAL_SLEEF_BUILD_SHARED_LIBS})
```
alexreinking, line 314:
```suggestion
  message(STATUS "Building shared libs : " ${BUILD_SHARED_LIBS})
```

Configure.cmake
```diff
@@ -7,7 +7,7 @@ include(CheckLanguage)
 
 if (SLEEF_BUILD_STATIC_TEST_BINS)
   set(CMAKE_FIND_LIBRARY_SUFFIXES ".a")
-  set(BUILD_SHARED_LIBS OFF)
+  set(_INTERNAL_SLEEF_BUILD_SHARED_LIBS OFF)
```
alexreinking, line 10:
```suggestion
  set(BUILD_SHARED_LIBS OFF)
```

Configure.cmake
```diff
@@ -834,7 +834,7 @@ endif()
 
 # Set common definitions
 
-if (NOT BUILD_SHARED_LIBS)
+if (NOT _INTERNAL_SLEEF_BUILD_SHARED_LIBS)
```
alexreinking, line 837:
```suggestion
if (NOT BUILD_SHARED_LIBS)
```

CMakeLists.txt
```diff
@@ -1,6 +1,5 @@
 
 # Options
-
 option(SLEEF_BUILD_STATIC_TEST_BINS "Build statically linked test executables" OFF)
```
alexreinking, line 3: The behavior of `option` and `set` are controlled by policies. You _must_ set `cmake_minimum_required` as the first line of your top-level program to have _any_ backwards compatibility guarantees at all.

Ideally `project()` would be the second line since _so many_ other commands depend on the internal state it configures.

```suggestion
cmake_minimum_required(VERSION 3.18)
project(SLEEF LANGUAGES C VERSION 3.6.0)

# TODO: rename uses of this variable to use the standard one
set(SLEEF_VERSION_PATCHLEVEL ${SLEEF_VERSION_PATCH})
set(SLEEF_SOVERSION ${SLEEF_VERSION_MAJOR})

option(SLEEF_BUILD_STATIC_TEST_BINS "Build statically linked test executables" OFF)
```

Then, after your options, add the following lines:
```
set(
if (SLEEF_ENABLE_CXX)
  enable_language(CXX)
endif ()
if (SLEEF_ENABLE_CUDA)
  enable_language(CUDA)
endif ()
```

No need to maintain that `LANGLIST` anymore. Just enable the requested languages.

Then, below, delete these lines:

```
cmake_minimum_required(VERSION 3.18)
if(${CMAKE_VERSION} VERSION_GREATER "3.14.99")
  cmake_policy(SET CMP0091 NEW)
endif()
```

`CMP0091` is enabled in CMake 3.15+ so the `cmake_policy` line is a noop and so the whole `if ()` block can be removed.

Also delete these lines:

```
set(SLEEF_VERSION_MAJOR 3)
set(SLEEF_VERSION_MINOR 6)
set(SLEEF_VERSION_PATCHLEVEL 0)
set(SLEEF_VERSION ${SLEEF_VERSION_MAJOR}.${SLEEF_VERSION_MINOR}.${SLEEF_VERSION_PATCHLEVEL})
set(SLEEF_SOVERSION ${SLEEF_VERSION_MAJOR})
set(LANGLIST C)
if (SLEEF_ENABLE_CUDA)
  set(LANGLIST ${LANGLIST} CUDA)
endif()
if (SLEEF_ENABLE_CXX)
  set(LANGLIST ${LANGLIST} CXX)
endif()
```

CMakeLists.txt
```diff
@@ -157,9 +163,9 @@ separate build directory. Note: Please remove autogenerated file \
 `CMakeCache.txt` and directory `CMakeFiles` in the current directory.")
```
alexreinking, line 163: Everywhere you should just use `SLEEF_SOURCE_DIR` and `SLEEF_BINARY_DIR` since those are the standard and expected names. The `project` command defines them and the `VERSION` variables.

https://cmake.org/cmake/help/v3.18/command/project.html#project

CMakeLists.txt
```diff
@@ -157,9 +163,9 @@ separate build directory. Note: Please remove autogenerated file \
...
 
-if(SLEEF_ENABLE_LTO AND BUILD_SHARED_LIBS)
-  message(FATAL_ERROR "SLEEF_ENABLE_LTO and BUILD_SHARED_LIBS cannot be specified at the same time")
-endif(SLEEF_ENABLE_LTO AND BUILD_SHARED_LIBS)
+if(SLEEF_ENABLE_LTO AND _INTERNAL_SLEEF_BUILD_SHARED_LIBS)
+  message(FATAL_ERROR "SLEEF_ENABLE_LTO and _INTERNAL_SLEEF_BUILD_SHARED_LIBS cannot be specified at the same time")
+endif(SLEEF_ENABLE_LTO AND _INTERNAL_SLEEF_BUILD_SHARED_LIBS)
 
 if(SLEEF_ENABLE_LTO)
   cmake_policy(SET CMP0069 NEW)
```
alexreinking, line 171:
```suggestion
```
That policy is already enabled by `cmake_minimum_required(VERSION 3.18)`. It's enabled for 3.9+.

docs/build-with-cmake.md
```diff
@@ -133,7 +133,7 @@ optimized, or any other special set of flags.
 - `SLEEF_SHOW_CONFIG` : Show relevant CMake variables upon configuring a build
 - `SLEEF_SHOW_ERROR_LOG` : Show the content of CMakeError.log
 
-- `BUILD_SHARED_LIBS` : Static libs are built if set to FALSE
+- `SLEEF_BUILD_SHARED_LIBS` : Static libs are built if set to FALSE
```
alexreinking, line 136:
```suggestion
- `SLEEF_BUILD_SHARED_LIBS` : Static libs are built if set to FALSE. If not manually set, then `BUILD_SHARED_LIBS` is used.
```

src/dft/CMakeLists.txt
```diff
@@ -373,8 +373,11 @@ foreach(T ${LIST_SUPPORTED_FPTYPE})
 endforeach()
 
 # Target libdft
-
-add_library(${TARGET_LIBDFT} $<TARGET_OBJECTS:dftcommon_obj> $<TARGET_OBJECTS:${TARGET_LIBARRAYMAP_OBJ}>)
+if(_INTERNAL_SLEEF_BUILD_SHARED_LIBS)
+  add_library(${TARGET_LIBDFT} SHARED $<TARGET_OBJECTS:dftcommon_obj> $<TARGET_OBJECTS:${TARGET_LIBARRAYMAP_OBJ}>)
+else()
+  add_library(${TARGET_LIBDFT} STATIC $<TARGET_OBJECTS:dftcommon_obj> $<TARGET_OBJECTS:${TARGET_LIBARRAYMAP_OBJ}>)
+endif()
```
alexreinking, line 380:
```suggestion
add_library(${TARGET_LIBDFT} $<TARGET_OBJECTS:dftcommon_obj> $<TARGET_OBJECTS:${TARGET_LIBARRAYMAP_OBJ}>)
```

src/libm/CMakeLists.txt
```diff
@@ -359,7 +359,12 @@ endif()
 # Original sleef sources
 set(STANDARD_SOURCES rempitab.c)
 
-add_library(${TARGET_LIBSLEEF} ${STANDARD_SOURCES})
+if(_INTERNAL_SLEEF_BUILD_SHARED_LIBS)
+  add_library(${TARGET_LIBSLEEF} SHARED ${STANDARD_SOURCES})
+else()
+  add_library(${TARGET_LIBSLEEF} STATIC ${STANDARD_SOURCES})
+endif()
```
alexreinking, line 366:
```suggestion
add_library(${TARGET_LIBSLEEF} ${STANDARD_SOURCES})
```
I'm going to stop highlighting all the places this needs to be reverted.

alexreinking (changes requested):
I got a little carried away with code review outside the question at hand; sorry about this. It's just that CMakeLists.txt that _don't_ start with `cmake_minimum_required`+`project` are liable to break without warning on any future version.

I don't understand how my blog post was used to justify splatting out `_INTERNAL_SLEEF_BUILD_SHARED_LIBS` everywhere. Just use `SLEEF_BUILD_SHARED_LIBS` as the value of `BUILD_SHARED_LIBS` if and only if it is defined by the user. I repeat the approach in my blog post and give an alternative approach.

alexreinking:
With the approach I outlined above, PyTorch should be able to build SLEEF by setting `SLEEF_BUILD_SHARED_LIBS` before calling `add_subdirectory`. I understand it is currently saving and restoring the value of `BUILD_SHARED_LIBS`. That will still work.

If you can show me with clear, reproducible steps in a self-contained post why my approach doesn't work, then I will be very happy to refine it further. I don't have time to read through 100+ comments across multiple PRs.

jgong5:
> With the approach I outlined above, PyTorch should be able to build SLEEF by setting `SLEEF_BUILD_SHARED_LIBS` before calling `add_subdirectory` (barring other bugs I'm not aware of). I understand it is currently saving and restoring the value of `BUILD_SHARED_LIBS`. That will still work.

Thanks @alexreinking for the suggestions. @blapie Sounds like this is exactly what we discussed in the meeting. I believe it should be doable.

xuhancn:
> I got a little carried away with code review outside the question at hand; sorry about this. It's just that CMakeLists.txt that _don't_ start with `cmake_minimum_required`+`project` are liable to break without warning on any future version.
> 
> I don't understand how my blog post was used to justify splatting out `_INTERNAL_SLEEF_BUILD_SHARED_LIBS` everywhere. Just use `SLEEF_BUILD_SHARED_LIBS` as the value of `BUILD_SHARED_LIBS` if and only if it is defined by the user. I repeat the approach in my blog post and give an alternative approach.

@alexreinking I use `_INTERNAL_SLEEF_BUILD_SHARED_LIBS` as an internal variable to control static/shared lib output. It is in order to avoid use `BUILD_SHARED_LIB` from the CMake offical document, snapshot as below:
image
`BUILD_SHARED_LIB` has too many limitation to use, I marked with highlight and red pen. I think my current PR is good enough and tested on pytorch.
If legacy user still use `BUILD_SHARED_LIBS` but not define `SLEEF_BUILD_SHARED_LIBS`, we can compatiable to them.
If new user defined `SLEEF_BUILD_SHARED_LIBS`, it will take over the control right.
Whatever, I think still use `BUILD_SHARED_LIBS` & `add_library` without `STATIC\SHARED` arg, which it not stronge enough and safe enough. Please agree to my current code, thanks. @alexreinking.

jgong5:
> > With the approach I outlined above, PyTorch should be able to build SLEEF by setting `SLEEF_BUILD_SHARED_LIBS` before calling `add_subdirectory` (barring other bugs I'm not aware of). I understand it is currently saving and restoring the value of `BUILD_SHARED_LIBS`. That will still work.
> 
> Thanks @alexreinking for the suggestions. @blapie Sounds like this is exactly what we discussed in the meeting. I believe it should work with PyTorch.

Oh, it seems I misunderstood the approach proposed by @alexreinking . Thanks @xuhancn for the reminder. I guess the difference is whether to overwrite the global `BUILD_SHARED_LIBS`. I doubt overwriting it would work (since PyTorch does the save/restore too and it doesn't work on Windows) and it is also not necessary. We only need to pass the SHARED/STATIC flag to the `add_library` according to `BUILD_SHARED_LIBS` and `SLEEF_BUILD_SHARED_LIBS` just like what @xuhancn without bothering altering the value of `BUILD_SHARED_LIBS`.

alexreinking:
@jgong5 what I suggested does **not** alter the global `BUILD_SHARED_LIBS` option. When you call `add_subdirectory` it creates a new variable environment. Setting a normal variable in a subproject is not visible to the parent project.

alexreinking:
@xuhancn - the documentation you cite outlines a pitfall of using `option` to do this. I do not use `option`. I use `set`.

`option` will create a cache variable when no variable at all exists. This will then be visible to the parent project on subsequent configure steps.

My solution does not suffer from this problem because it does not touch the cache.

xuhancn:
> @xuhancn - the documentation you cite outlines a pitfall of using `option` to do this. I do not use `option`. I use `set`.
> 
> `option` will create a cache variable when no variable at all exists. This will then be visible to the parent project on subsequent configure steps.
> 
> My solution does not suffer from this problem because it does not touch the cache.

Currently, we need to guarantee it works on pytorch's submodule. My PR code is validated on pytorch all PreCI.
image
If you persist in your solution and willing to contribute to sleef. Could you please submit your solution as a sleef PR? I will help on validate your PR on pytorch.

alexreinking:
> Instead I suggested a simple alternative, that you don't want to try, which is encapsulating the options and `add_subdirectory(sleef)` in a function. I did it for you and it works, as far as your simple example goes, it does not modify `BUILD_SHARED_LIBS` in the global scope 

@blapie -- I think you're 100% correct in your assessment that PyTorch _could_ use a function to set up `BUILD_SHARED_LIBS` before calling `add_subdirectory`. I still think that providing `SLEEF_BUILD_SHARED_LIBS` is _nicer_ for downstreams because doing that function dance is pretty ugly. The situation has gotten a little better since `block()` was introduced in CMake 3.25.

xuhancn:
Currently,we have two PRs can fix `BUILD_SHARED_LIB` unexpected behavior on Windows MSVC.
1. This PR: https://github.com/shibatch/sleef/pull/510
2. alexreinking's PR: https://github.com/shibatch/sleef/pull/531
Both of them can pass pytorch's PreCI.

| |[PR_#510](https://github.com/shibatch/sleef/pull/510)|[PR_#531](https://github.com/shibatch/sleef/pull/531)|
|----|----|----|
|Fix build issue| Yes | Yes |
|Implement| Avoid to use `BUILD_SHARED_LIB` | Overwrite to `BUILD_SHARED_LIB`|
|Support `SLEEF_BUILD_SHARED_LIB`| Yes | Yes |
|compatible to `BUILD_SHARED_LIB`| Yes | Yes |
|Easy to maintain| Basic CMake knowledge is ok | Need deep dive to CMake|

I made a comparsion table to summary the two PRs. My PR is simpler than alexreinking's. My suggestion is merge my PR into master branch.

Any comments? @blapie @jgong5 @alexreinking

jgong5:
> I made a comparsion table to summary the two PRs. My PR is simpler than alexreinking's. My suggestion is merge my PR into master branch.
> 
> Any comments? @blapie @jgong5 @alexreinking

I don't have strong preference on the option. Both options LGTM.

alexreinking:
@xuhancn - I think your comparison is wildly unfair.

> My PR is simpler than alexreinking's.

My PR consists of _three_ lines that matter for the shared libraries issue. It also fixes a real misuse of CMake and improves some console output.

Your PR requires ongoing maintenance to ensure that any new `add_library` call is guarded by a check to the unnecessary and nonstandard variable you introduced.

With my PR, nothing special needs to be done going forward.

> Need deep dive to CMake

This is also untrue. One just needs basic facts about how variable scope works in CMake.

---

Re: @jgong5 
Cc: @blapie 

> I don't have strong preference on the option. Both options LGTM.

I encourage the maintainers here to look at both diffs and draw their own conclusions.

blapie:
@alexreinking - Hello! Thanks a lot for the insight, I learned a lot reading through your suggestion. First of all, I agree that SLEEF's cmake was in need for some proper housekeeping, which made it prone to bugs.
Thank you for making it clear where the issue most likely came from (set/option, overwriting to cache, ...). I agree that the `SLEEF_`-prefixed variable can be helpful for downstream projects, and I like that we stick to using the standard variable internally.
I am totally conscious that you did not have to respond to our enquiry, let alone provide so much insight as well as an entire PR/solution. I wish your contribution was welcomed with a bit more warmth...

@jgong5 - I believe this solution meets in the middle while make SLEEF build system way more robust. Thanks for contributing to the discussion and sorry I could not reply before.

I will review #531 but from the first couple of messages it sounded all good.

xuhancn:
SLEEF will merge this PR(https://github.com/shibatch/sleef/pull/531) to fix the build issue. Let's wait for a while.

blapie:
Happy we finally got to a satisfying solution for everyone, thanks @alexreinking, @xuhancn and @jgong5.

jgong5:
@blapie @alexreinking Thanks a lot for your help and effort for landing it!

PR #515, opened 2024-02-12
Fix SLEEF_BUILD_SCALAR_LIB lost function

xuhancn:
Fix "SLEEF_BUILD_SCALAR_LIB" lost function to control build sleefscalar lib.

|Status|Control Build|Control Install|
|----|----|----|
|Current Code| No |Yes|
|This PR|Yes|Yes

Original code: `SLEEF_BUILD_SCALAR_LIB` only control whether install scalar_lib, it would always build scalar_lib.
My PR: `SLEEF_BUILD_SCALAR_LIB` control both whether build and install scalar_lib.

### DIFF
--- a/src/libm/CMakeLists.txt
+++ b/src/libm/CMakeLists.txt
@@ -972,25 +972,24 @@ endif()
 # TARGET_LIBSLEEFSCALAR
 # --------------------------------------------------------------------
 # Build scalar-only library from sleefdp.c and sleefsp.c
-
-add_library(sleefscalar sleefdp.c sleefsp.c rempitab.c)
-add_dependencies(sleefscalar ${TARGET_HEADERS})
-set_target_properties(sleefscalar PROPERTIES
-  VERSION ${SLEEF_VERSION}
-  SOVERSION ${SLEEF_SOVERSION}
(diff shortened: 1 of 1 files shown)


blapie:
Hi! Could you please give a bit of context? 

What do you mean by lost function? Do you mean that library is build but not installed?

I can see a benefit in choosing to build scalar libs or not, especially if they are not installed after all.

Just a quick reminder. In the original design, scalar routines are not meant to be used but rather here as a reference for vector implementations. So it is recommended not to use them when performance matters.

xuhancn:
> Hi! Could you please give a bit of context?
> 
> What do you mean by lost function? Do you mean that library is build but not installed?
> 
> I can see a benefit in choosing to build scalar libs or not, especially if they are not installed after all.
> 
> Just a quick reminder. In the original design, scalar routines are not meant to be used but rather here as a reference for vector implementations. So it is recommended not to use them when performance matters.

Original code: `SLEEF_BUILD_SCALAR_LIB` only control whether install scalar_lib, it would build it always.
My PR: `SLEEF_BUILD_SCALAR_LIB` control both whether build and install scalar_lib.

update to description also.

PR #520, opened 2024-02-16
Reconcile RVV mask types.

sh1boot:
Make the vmask types for single-precision and double-precision interfaces compatible, so that we have fewer divergent functions.

Keep the vopmask types incompatible, because the underlying vbool*_t types still encode some underlying assumptions about the number of lanes involved.

Make the still-incompatible function names distinct between SP and DP, and then reconcile these with the preprocessor at the end of the file so that the caller still sees the expected operations.  This will help when it comes to generating inline headers in a future revision.

### DIFF
--- a/src/arch/helperrvv.h
+++ b/src/arch/helperrvv.h
@@ -56,43 +56,6 @@ static INLINE int vavailability_i(int name) { return -1; }
 // RISC-V Vector Types
 ////////////////////////////////////////////////////////////////////////////////
 
-// About the RISC-V Vector type translations:
-//
-// Because the single- and double-precision versions of the RVV port have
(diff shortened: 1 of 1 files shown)


sh1boot:
This contains a portion of #503, and fits underneath that PR, so that each can be reviewed in isolation.

luhenry (approved):
And all tests pass on CI: https://github.com/rivosinc/sleef/actions/runs/7925708693

cc @ericlove @GlassOfWhiskey for review

luhenry:
Hi @blapie, what could we do to get this PR (and associated ones https://github.com/shibatch/sleef/pull/521 https://github.com/shibatch/sleef/pull/522) to go forward?

PR #522, opened 2024-02-16
Add RVV inline header support.

sh1boot:

Issue #524, opened 2024-02-17
RISC-V architecture missing on https://sleef.org/


xypron:
With the 3.6 release support for the RISC-V architecture has been added. But it does not show up in the table in https://sleef.org/index.xhtml#environment.

According to toolchains/riscv64-*.cmake clang >= 17 or gcc >= 14 is needed. This should also be mentioned in this section.

Best regards

Heinrich

blapie:
Hello! Thanks for reporting that. It is definitely our omission. However the plan is to remove this table in the near future, since some of these features have retrograded to experimental. But it would make sense for it to contain Risc-V, until actually replaced by the table in README.

By the way, ultimately we would like to switch from html to markdown sources and let github pages render the website.

PR #527, opened 2024-03-04
add comments to helperrvv.h

sh1boot:
Depends on #520, and #521, #503, and #522.

The first commit is already in the #521 PR, but repeated here so I don't have to rebase #503 and #522.
The relevant changes are: https://github.com/rivosinc/sleef/commit/78ea02c2ab1df54d929a92fcaf5935e06c714563

### DIFF
--- a/src/arch/helperrvv.h
+++ b/src/arch/helperrvv.h
@@ -71,6 +71,12 @@
 #endif
 
 #if __riscv_v_intrinsic <= 12000
+// __riscv_vcreate* intrinsics only showed up in v1.0-rc0 of the RVV intrinsics
+// spec and have already been implemented in clang-18, but are useful for
+// eliminating issues with uninitialised data because they are explicit that
+// the whole result has defined values.  Here we do our best to offer fallback
+// implementations where needed.
+//
 #define __riscv_vcreate_v_f32m1_f32m2(x, y) __riscv_vset(__riscv_vlmul_ext_v_f32m1_f32m2(x), 1, y)
 #define __riscv_vcreate_v_f32m2_f32m4(x, y) __riscv_vset(__riscv_vlmul_ext_v_f32m2_f32m4(x), 1, y)
 #define __riscv_vcreate_v_f32m4_f32m8(x, y) __riscv_vset(__riscv_vlmul_ext_v_f32m4_f32m8(x), 1, y)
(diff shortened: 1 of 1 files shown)


luhenry:
@blapie rebased this one as well given all the previous PRs were merged.

src/arch/helperrvv.h
```diff
@@ -1292,7 +1335,13 @@ static int vcast_i_vi2(vint2 v) {
 //
 
 static vquad loadu_vq_p(const int32_t *ptr) {
-    return SLEEF_RVV_DP_VREINTERPRET_VQ(SLEEF_RVV_DP_VREINTERPRET_4VU(SLEEF_RVV_SP_LOAD_2VI(ptr, VECTLENSP * 2)));
```
luhenry, line 1295: This is only removing a trailing space but doesn't change the code.

luhenry (approved):

src/arch/helperrvv.h
```diff
@@ -71,6 +71,11 @@
 #endif
 
 #if __riscv_v_intrinsic <= 12000
+// __riscv_vcreate* intrinsics only showed up fairly recently, but are useful
```
blapie, line 74: Can we give be a little bit more specific than "fairly recently"?

src/arch/helperrvv.h
```diff
@@ -71,6 +71,11 @@
...
+// for eliminating issues with uninitialised data because they are explict that
```
blapie, line 75: explicit

src/arch/helperrvv.h
```diff
@@ -920,6 +950,10 @@ static INLINE vint vcastu_vi_vm(vmask vm) {
 static INLINE vint vcast_vi_vm(vmask vm) {
   return SLEEF_RVV_DP_VREINTERPRET_VI(__riscv_vncvt_x(vm, VECTLENDP));
 }
+
+// Theese are the complementary case to the earlier comment about
```
blapie, line 954: These

blapie (commented):
Just pointing out a few typos. Otherwise lgtm

blapie:
I will take care of the general documentation on the website and README (as part of #524), but could you please provide a reference file like https://sleef.org/aarch64.xhtml for rvv types and functions.

src/arch/helperrvv.h
sh1boot: I've updated the comment but it turns out I'll need to update the logic as well.

It turns out it was only a coincidence that the feature added at the same time as the feature test macro changed in the compiler versions I checked.  Clang took a double-step, updating to 0.12 and then progresively adding the draft features without a version bump.  GCC is stuck in between, advertising 0.12 but not implementing vcreate (which is correct, I am wrong).

I'll submit a separate patch to test for clang-18 and up, as well as intrinsics version 1.0 and up.

sh1boot:
> but could you please provide a reference file like https://sleef.org/aarch64.xhtml for rvv types and functions.

It's going to take me a while to find time for this (several weeks, at least).  AIUI I can broadly copy-paste the SVE documentation but I'll also need to go through the list and make sure everything is present and true.

PR #530, opened 2024-03-12
Fix RVV intrinsic version detection.

sh1boot:
__riscv_vcreate* appear after RVV intrinsics v0.12, and the next version after that is v1.0 which is not yet ratified.

Nevertheless, clang-18 and up already have the support we need.

### DIFF
--- a/src/arch/helperrvv.h
+++ b/src/arch/helperrvv.h
@@ -70,7 +70,7 @@
 #endif
 #endif
 
-#if __riscv_v_intrinsic <= 12000
+#if __riscv_v_intrinsic < 1000000 && !(defined(__clang_major__) && __clang_major__ >= 18)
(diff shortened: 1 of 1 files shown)


sh1boot:
Again with this chaining thing... this PR contains #527 and will need to be rebased after that goes in.  Meanwhile the diff is just the one patch:
https://github.com/shibatch/sleef/commit/82dc7a2041146b14ddf99c968e07c277030902dd

blapie:
This make sense to me, will merge after rebase. Wondering if we should consider testing with clang 18 soon, unless this introduces complications.

sh1boot:
Clang-18 has only just been released, I think (at 18.1 with no 18.0?).  I'm not sure how long it will take for all the tooling to propagate.

I've been testing locally using apt.llvm.org snapshots.  It should be good to go.

PR #536, opened 2024-04-15
Mark Sleef_rempitabq* constants as static

luhenry:
Without marking it static, it fails to compile when including multiple sleefinline_*.h into the same compilation unit.

Co-Authored-By: Hamlin Li <hamlin@rivosinc.com>


Hamlin-Li (approved):
Looks good. Thanks

joeramsay:
Would you mind adding static to `Sleef_rempitabsp` as well?
> including multiple sleefinline_*.h into the same compilation unit.

This should already be addressed by the include guard around the rempi tables. Are you sure the issue isn't from linking multiple TUs that all included the header and used the table?

luhenry:
@joeramsay fixed.

luhenry:
@blapie hi, what would be the path forward for this? Thanks!

nSircombe:
Hi @luhenry, it should be possible to move this on soon (see https://github.com/shibatch/sleef/pull/537#issuecomment-2063895882)

PR #539, opened 2024-05-16
Switch to gcc14 in GHA

joanaxcruz:
In GCC12 and GCC13 latest versions, the bug which was causing test failures in this project has been fixed. In GCC14 release series, the bug is not present. Hence we transition our testing into the latest compilers.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  GCC_VERSION: "11"
+  GCC_VERSION: "14"
   LLVM_VERSION: "17"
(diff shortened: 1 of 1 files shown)


joanaxcruz:
GCC-14 (and G++-14) is only available in the Universe repository for Ubuntu 24.04, as evident in the [Ubuntu Package archive](https://packages.ubuntu.com/noble/gcc-14).
However, GHA does not support  Ubuntu24.04 runner's yet (only available as a Beta for now https://github.blog/changelog/2024-05-14-github-hosted-runners-public-beta-of-ubuntu-24-04-is-now-available/)
Hence, I will try completing this PR when these runners are officially released (predicted for Q2).

blapie:
Superseded by #601

Issue #320, opened 2020-08-10
16-bit floating point support


alexeyr:
It's mentioned in https://github.com/shibatch/sleef/issues/227#issuecomment-432280602 but I wanted to make a separate issue explicitly asking for FP16 support; as mentioned, it's very useful for machine learning.

shibatch:
FP16 is a bit different from other precisions since a single table lookup can implement any function. There are also two kinds of FP16.
Which functions do you need for machine learning? My understanding is that there are many special functions that are only used in machine learning, so I think it is better to implement those functions directly.

alexeyr:
GELU in particular (`0.5 * x * (1 + erf(x / sqrt(2)))`).

shibatch:
I am now looking into this thing, but it seems not so many kinds of hardware support bfloat16 of IEEE 754 half-precision operations. AVX-512 BF16 only does conversion and dot product.
Which specific hardware do you need to be supported?
Only those three operators? Or is it better to create basic operators like add, mul, div, etc?

shibatch:
I made a prototype of table-lookup-based implementation of a vectorized unary function. This implementation lacks rounding and other features.

https://gist.github.com/shibatch/42f82c885f62bc2c063d6500844f937f

Please let me know if things like this is what you need.

alexeyr:
> Which specific hardware do you need to be supported?

We only care about NEON in this particular project. Though it may get shut down before we get to adding FP16 support anyway now :(

shibatch:
It is not likely that I can achieve significant speedup with neon, because it does not have either bfloat16 support or gather operator.

alexeyr:
Unfortunately, now we definitely won't be adding FP16 in any near term. The issue can be closed unless you'd like to keep it open.

shibatch:
Okay. I am now planning to add a FP16 support regardless of what you need.
Thank you for your suggestion.

shibatch:
Hello,

What kind of functionality do you need?
There are two problems in the actual development process: the problem of deciding what the specifications will be and the uncertainty as to whether the implementation of such a function will be fast enough to perform the calculation.
In tlfloat, I implemented fp16, but this is because tlfloat implements all functions with generic precision, and fp16 was implemented as part of that, so in that sense, I implemented the functionality without considering the computation speed.
[...]

PR #540, opened 2024-05-18
Allow build dft on MinGW

Andarwinux:
It actually works fine.


blapie:
Hello! We are a bit concerned about the absence of testing for this one. But I suppose DFT can still be disabled manually.
Any chance you could test with gcc?

blapie:
Btw thanks for testing and reporting this, very useful! I'll try to test locally.

blapie:
Thanks for the contribution, impressive GHA skills!

I like the idea of the build/test-msys2 jobs, we were definitely missing testing for Windows (x86).

On the other hand, we did not desperately need a precommit workflow for llvm-mingw, but it is quite handy to be able to (cross-)compile for windows on both x86 and arm64, so happy try and merge that too.

Before getting into the details, would you mind moving both jobs to separate yaml files (even if it means duplicating code for build-native)? We might want to schedule and parametrise the 3 workflows differently. `{build,test}-msys2` to `build-and-test-windows.yml` and `build-llvm-mingw` to `build-cross-windows.yml`?

Btw if you are aware of simple/nice ways to re-use bits of workflows (say build-native) in other workflows, please let us know.

blapie:
> One possible way to do this is to save native builds to action cache, which is quite a bit faster than artifacts and can across workflows. But I suspect this would still be slower than build native and cross within same instance.

Thanks for the insight, action cache might be useful for other jobs. I like the present solution though, hopefully in the future a native build isn't even needed to cross-compile.

blapie:
Regarding the libm and quad tests on msys2, would you be able to link us to the build or test failures?
The main reason I expect the build to fail is that wait.h is not available on msys2, since it is POSIX. For this reason I think the suggested changes to CMakeLists are acceptable, until we find a workaround.

Do you have suggestions on ways to port a code that relies on wait.h (wait, fork, ...) to mingw/msys2?

@shibatch Have you ever considered ways to make the libm tester compatible with mingw?

shibatch:
Since tester1 does not need to be compiled with mingw, the test should be possible by combining tester1 for Cygwin or linux, although it may be a bit troublesome to write the script for the test.

Another idea could be to add a new tester using tlfloat, which I am currently developing. With tlfloat, testing is possible even if libmpfr is not available.

blapie:
> Since tester1 does not need to be compiled with mingw, the test should be possible by combining tester1 for Cygwin or linux, although it may be a bit troublesome to write the script for the test.

Fair, it's true that they can be compiled independently, but I'm afraid this will confuse things more than they already are on Windows.

> Another idea could be to add a new tester using tlfloat, which I am currently developing. With tlfloat, testing is possible even if libmpfr is not available.

That could be interesting! Looks like you already have support for several OS-es there too. Mpfr is not the worst dependency to have, but always good to remove dependencies I suppose.

Would tlfloat resolve the dependency on wait.h and POSIX in general? Is it able to handle multithreading too?
My suggestion would have been to just brutally remove multithreading/calls to wait.h in tester1 on windows, but I'm not sure how to do that.

Alternatively, we could look at straight windows port for wait.h https://github.com/win32ports/sys_wait_h ?

shibatch:
The new tester can be made pipe-free and multi-threading capable.

blapie:
> The new tester can be made pipe-free and multi-threading capable.

I suggest we create an issue specifically for this?

shibatch:
Writing an Issue is fine, but there are other things to consider regarding this.
First is how to link tlfloat from sleef.

If we implement the fp16 support mentioned here, it could be related to that as well.
https://github.com/shibatch/sleef/issues/320

.github/workflows/build-and-test-windows.yml
```diff
@@ -0,0 +1,149 @@
...
+      - name: Upload build-msys2-${{ matrix.sys }} artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: build-msys2-${{ matrix.sys }}
+          path: |
+            _build-*
+            _install-*
+        if: always()
+
+  test-msys2:
```
blapie, line 88: Could we display information on the cpu in this job at least? Equivalent to what we have in the linux workflow:
https://github.com/shibatch/sleef/actions/runs/8845094575/workflow#L102
Ideally in all jobs but that might be a bit much.

.github/workflows/build-cross-windows.yml
```diff
@@ -0,0 +1,108 @@
...
+
+          cmake -S . -B _build-${{ matrix.arch }} -GNinja \
+            -DCMAKE_INSTALL_PREFIX="$(pwd)/_install-${{ matrix.arch }}" \
+            -DCMAKE_SYSROOT="$(pwd)/llvm-mingw/${{ matrix.arch }}-w64-mingw32" \
+            -DNATIVE_BUILD_DIR="$(pwd)/_build-native" \
+            ${COMMON_CMAKE_FLAGS} \
+            -DCMAKE_C_COMPILER=${{ matrix.arch }}-w64-mingw32-clang \
+            -DCMAKE_ASM_COMPILER=${{ matrix.arch }}-w64-mingw32-clang \
+            -DCMAKE_SYSTEM_NAME=Windows \
+            -DCMAKE_SYSTEM_PROCESSOR=${{ matrix.arch }} \
```
blapie, line 97: Do you think we can rely on a toolchain file for most of these options? Our toolchain files are located in https://github.com/shibatch/sleef/tree/master/toolchains.

.github/workflows/build-and-test-windows.yml
```diff
@@ -0,0 +1,149 @@
...
+jobs:
+  build-msys2:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sys:
+          - clang64
+          #- mingw64
+          #- ucrt64
```
blapie, line 37: Could we have a quick comment saying why these are commented out?

.github/workflows/build-cross-windows.yml
```diff
@@ -0,0 +1,108 @@
...
+            EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DSLEEF_ENFORCE_SSE2=ON -DSLEEF_ENFORCE_SSE4=ON -DSLEEF_ENFORCE_AVX=ON -DSLEEF_ENFORCE_AVX=ON -DSLEEF_ENFORCE_AVX2=ON -DSLEEF_ENFORCE_AVX512F=ON -DSLEEF_ENFORCE_FMA4=ON"
+          fi
+
+          cmake -S . -B _build-${{ matrix.arch }} -GNinja \
+            -DCMAKE_INSTALL_PREFIX="$(pwd)/_install-${{ matrix.arch }}" \
+            -DCMAKE_SYSROOT="$(pwd)/llvm-mingw/${{ matrix.arch }}-w64-mingw32" \
+            -DNATIVE_BUILD_DIR="$(pwd)/_build-native" \
+            ${COMMON_CMAKE_FLAGS} \
+            -DCMAKE_C_COMPILER=${{ matrix.arch }}-w64-mingw32-clang \
+            -DCMAKE_ASM_COMPILER=${{ matrix.arch }}-w64-mingw32-clang \
```
blapie, line 95: `CMAKE_ASM_COMPILER` one does not seem to be used, could just rm?

.github/workflows/build-and-test-windows.yml
blapie: Do you have any more details on that, maybe a link? Is the compiler broken or is it SLEEF when compiled in a gcc envs?

README.md
```diff
@@ -33,54 +33,54 @@ The following table summarises currently supported vector extensions, compilers
...
   <th>gcc</th><th>llvm</th><th>icc</th>
   <th>gcc</th><th>llvm</th>
-  <th>gcc</th><th>llvm</th><th>msvc</th>
+  <th>gcc</th><th>llvm-gnu</th><th>llvm-msvc</th><th>msvc</th>
 </tr>
 <tr align="center"><th>x86_64</th><th>SSE2, SSE4,<br>AVX, AVX2, AVX512F</th>
   <td>:green_circle:</td><td>:green_circle:</td><td>:white_circle:</td>
   <td>:white_circle:</td><td>:white_circle:</td>
-  <td>:white_circle:</td><td>:white_circle:</td><td>:white_circle:</td>
+  <td>:white_circle:</td><td>:green_circle:</td><td>:white_circle:</td><td>:white_circle:</td>
```
blapie, line 53: Thanks for taking the time to document that.
In this case only DFT is tested though. What we do at the moment is at least add a comment in the following subsections, here `Component support` section should indicate that on Windows only DFT is tested but everything is built.

Let's not bother in this patch but we should probably use a different colour scheme when only part of the library is tested, or when component is compiled only and not tested. Ultimately, this could probably be mapped to the workflow/job status.

.github/workflows/build-and-test-windows.yml
```diff
@@ -0,0 +1,160 @@
...
+      - uses: actions/checkout@v4.1.1
+        with:
+          persist-credentials: false
+
+      - name: Build msys2-${{ matrix.sys }}
+        shell: msys2 {0}
+        run: |
+          EXTRA_CMAKE_FLAGS=""
+          if [[ ${{ matrix.sys }} = "clang64" ]]; then
+            EXTRA_CMAKE_FLAGS="$EXTRA_CMAKE_FLAGS -DSLEEF_ENABLE_LTO=ON"
```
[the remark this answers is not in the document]
blapie: Ok, works for me. But please refrain from adding too many variants/parameters as we have limited resources to address failures. Also special casing for combinations of parameters is a bit gross in GHA workflow files.

blapie:
Are you happy for me to merge?

PR #563, opened 2024-07-18
Add mutex in feature detection code

Andreas-Krebbel:
This implements a suggestion from @shibatch for #561. It uses a pthread mutex to prevent multiple threads from accessing the sigjmp variable in the feature detection code simultaneously.

Remaining todos:
1. How to initialize the pthread mutex on MacOX and Windows? The code currently errors out on these platforms
2. With older Glibc's I guess this requires sleef users to link against libpthread. Would that be acceptable? With newer Glibcs libpthread is part of libc already.

Apart from that the code fixes the problems in #561

### DIFF
--- a/src/libm/dispatcher.h
+++ b/src/libm/dispatcher.h
@@ -19,16 +19,46 @@ static sigjmp_buf sigjmp;
 #define LONGJMP siglongjmp
 #endif
 
+// Protect sigjmp from parallel access
+// FIXME: Static initializers will not work on Windows and MacOS
+#ifdef __linux__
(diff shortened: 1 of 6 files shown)


shibatch:
On Windows, you can use InitOnceExecuteOnce to initialize the mutex.
Or you could call the entire detection code from InitOnceExecuteOnce.
I think we do not need to write this code now.

https://devblogs.microsoft.com/oldnewthing/20110408-01/?p=10983
https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-initonceexecuteonce

I see no problem linking libpthread.

PR #564, opened 2024-07-30
Port documentation from html to markdown

blapie:
This allow rendering the website on both Github UI and a web page using a customisable theme ([just-the-docs](https://github.com/just-the-docs/just-the-docs)),while simplifying maintenance.

The theme is actively maintained and offer efficient navigation, a search engine, and a modern layout that display nicely on laptop, phones and tablets.

[Test page can be previewed here](https://blapie.github.io/test-page/)

In order to separate docs from the content we need to add a Gemfile (for pluggins) and GHA workflow (for deploying website). Thus it is not as simple as adding a `_config.yml`, but it could be improved in the future.

Syntax highlighting can be customised later.

Some light updates of the docs but content is mostly untouched.


blapie:
Hello @shibatch! This PR is for updating the layout of the website while facilitating maintenance work. Bonus is that the website renders in Github.

Since it is quite a substantial update, but also difficult to review, we wanted to give you a heads up, to see what you think.

You can see it tested inside a test project https://github.com/blapie/test-page/tree/main/docs
and see it live here https://blapie.github.io/test-page/

We have reviewed this internally, so we know the website works.

Still need to integrate the google verification code in to the config and fix up the favicon.

blapie:
Hi @shibatch, In order for us to merge this PR could you please go to https://github.com/shibatch/sleef/settings/pages, and change the Build & Deployment source to Github Actions, if that's ok? That will allow the new custom workflow to deploy the page.

shibatch:
I missed mentioning this because I didn't expect it to be completed so quickly, but how do you think about the idea of adopting asciidoc?
I think markdown is fine for a simple structured document, but asciidoc would be better for a document of this scale.
It probably won't take much time to rewrite.
It can also render on github.
I'm not saying that asciidoc is absolutely necessary, but could you please give it some consideration?

blapie:
hm, interesting! I have to admit I am not familiar with asciidoc. Looking at side by side comparison I do think they both offer what we need, while asciidoc maybe offers slightly more customisation.

My preference would still go to markdown as it seems ubiquitous, I cannot really name a project I know that uses asciidoc.

Probably my main arguments against asciidoc would be that contributors/maintainers might be more familiar with asciidoc, and also that I am unsure how well it is supported by most github page themes.

Visually I think my heart still goes to plain MD versus plain AD, but that is just a matter of taste.

I'm also worried of some limitations we may find further down the line, which always happens with this kind of big porting tasks.

I believe MD is good enough and simple enough. If more customisation is required, I think it is still acceptable to use plain html. In a way I like how MD forces us to keep the docs simple, in not allowing too much customisation.

Finally, most of the porting job is done in this patch anyway. As you said moving from MD to AD, could be a subsequent task, that's relatively simple to achieve.

shibatch:
@brapie I changed Build & Deployment source to Github Actions. Please confirm.

I'll take a quick look at the page you created and write down my thoughts.

First of all, do we need to continue to display all partners in general? NAIST is at the top of the list because I used to belong to them, but they were very unsympathetic to my contributions to SLEEF, so I think NAIST can be erased now.

RISC-V related contributions are substantial, so I think you should write better about them.

In the "Compiling the library with Microsoft Visual C++" section in the User guide, please remove "&amp;"

If you are going to change the left menu to "Extra", please change the headings in the main text to that as well. Please be careful to maintain consistency. 


I will write again if I notice anything else.

blapie:
Thanks for the change and the feedback!

> First of all, do we need to continue to display all partners in general? NAIST is at the top of the list because I used to belong to them, but they were very unsympathetic to my contributions to SLEEF, so I think NAIST can be erased now.

Rm-ed logo. Is there another institution you should be affiliated with in the docs? I updated to "formerly at NAIST".

> RISC-V related contributions are substantial, so I think you should write better about them.

Sure, done!

> In the "Compiling the library with Microsoft Visual C++" section in the User guide, please remove "&"

Done.

> If you are going to change the left menu to "Extra", please change the headings in the main text to that as well. Please be careful to maintain consistency.

Absolutely. Done.

For further docs update I'll leave it to another PR, but please feel free to submit a doc update issue, and Ill make the changes.
Thanks again!

shibatch:
Something is wrong with Sleef_floorf.

https://blapie.github.io/test-page/2-references/libm/#sleef_floorf

blapie:
Good spot. Fixed!

Issue #574, opened 2024-09-16
Add a new tester that utilizes tlfloat


shibatch:
**Why a new feature? Is your feature request related to a problem? Please describe it.**

- The new tester will work without a pipe connecting two processes. Introducing this tester will resolve the posix dependency problem.

- The new tester can be used in environments where MPFR is not available. It will be much easier to do testing on Windows.

- I expect the new tester to work faster than the current tester/iut combo.

**What feature would you like?**

- The tester compares the output of SLEEF to the output of [TLFloat](https://github.com/shibatch/tlfloat) for a predetermined set of inputs.
- By comparing the outputs to the higher precision output of TLFloat, the tester verifies that the error in SLEEF's outputs is within a predetermined range.

**How would you like this feature to be implemented?**

- As described in #568, I will incorporate TLFloat as a submodule of SLEEF.

- The structure will be like a combination of the existing IUT and tester, and will be built with different settings for each vector extension.

**How portable is the feature across architectures and platforms?**

- The purpose of introducing this tester is to address the current portability issues associated with MPFR and pipe.

- TLFloat has fewer portability issues than MPFR and can be used in a wider range of environments.

- However, it will require a C++20 compiler and cannot be compiled with clang 16 or earlier due to compiler bugs.

**Describe alternative approaches you have considered**

- The current tester/iut will not be deleted.

- At the time of build, the user can select either the current tester/iut or the new tester to be introduced.

- I will allow the user to choose whether to build the current tester/iut and the new tester by cmake options.

**Additional context**

- The addition of the new tester will not constitute a major code change.

- However, the addition of C++ code and the introduction of submodules may affect the user's build environment and future development.

- The default setting is to test with the existing tester/iut, as is the case at present. Only when new cmake options are specified, the new tester will be built and used for testing instead of the current tester/iut.

blapie:
Hello, Sorry for the delay in replying. 

Here are a few general comments:

1. This generally seems like an improvement and the integration of the new tester should be fairly smooth.

2. As mentioned in the previous issue, the main worry is the addition of code, which means more maintenance work. However, because it solves a portability issue (in particular the posix-one) then I think it is worth the maintenance cost.

4. I believe the project can greatly benefit from C++ in the long term (e.g. leverage templatisation to reduce the need for macros), in particular when it comes to tests and benchmarks. By the way we are currently reworking the benchmark suite using C++. 

5. I think the existing testers could use a bit of documentation (inline and on website) to facilitate maintenance.

[...]

PR #586, opened 2024-10-10
Update the helper files so that the library can be compiled with newer compilers

shibatch:
# Checklist

* This update is required to build the library with newer compilers, especially MSVC.

# What changes did you make?

* Notation like ((vdouble) { +0.0, -0.0, +0.0, -0.0 }) is not compliant to the standards, and had to be replaced with vector-set intrinsics.
* Added explicit typecasting so that the code can be compiled with C++ which has strict type checking.


blapie:
Sorry, I was just testing the new compatibility checks, if you rebase your change these 2 checks should pass now.

PR #587, opened 2024-10-10
Update the helper files so that the library can be compiled with newer compilers

shibatch:
# Checklist

* This update is required to build the library with newer compilers, especially MSVC.

# What changes did you make?

* Notation like ((vdouble) { +0.0, -0.0, +0.0, -0.0 }) is not compliant to the standards, and had to be replaced with vector-set intrinsics.
* Added explicit typecasting so that the code can be compiled with C++ which has strict type checking.

This is manually rebased https://github.com/shibatch/sleef/pull/586.

### DIFF
--- a/src/arch/helperavx.h
+++ b/src/arch/helperavx.h
@@ -110,7 +110,7 @@ static INLINE int vavailability_i(int name) {
 
 #endif // #if !defined(SLEEF_GENHEADER)
 
-static INLINE void vprefetch_v_p(const void *ptr) { _mm_prefetch(ptr, _MM_HINT_T0); }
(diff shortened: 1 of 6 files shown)


blapie:
Looks good to me!

Not required for this patch but to claim support it would be good if we could build (and test) with MSVC in GitHub Actions. Created #588 for that.

Btw: I think you could have rebased without creating a new PR (sync fork with upstream, then rebase upstream/master onto existing branch, then force push branch to your fork).

shibatch:
I know I can rebase without creating a new PR, but I am unfamiliar with git and wasn't confident I could do it right. I did this because if I rewrite the main branch by mistake, it could be quite messy.

PR #589, opened 2024-10-10
Integrate Google benchmarks into SLEEF

joanaxcruz:
* Improve code quality 
* Documentation update

# What changes did you make?
Added new benchmark tool to SLEEF project using googlebench framework.
In theory this tool can benchmark any unary and binary function in SLEEF.
Benchmark is enabled in all functions listed in benchsleef.cpp. This list uses macros in benchmark_callers.cpp, so that we can enable benchmarking in multiple precisions using single lines of code. It is also possible to list the exact function we want, as each of the macros in benchmark_callers.cpp can be called independently and connected like building blocks.
The tool is integrated with SLEEF via CMake, meaning it can be built automatically when SLEEF is built. To enable that, pass CMake argument -DSLEEF_BUILD_BENCH=ON.

This tool depends on C++17 standard.
Tested on aarch64 for scalar, vector and SVE routines Tested on x86 for different vector length extensions. Tested for llvm-17, gcc-11 and gcc-14.

This PR relates to https://github.com/shibatch/sleef/issues/9.
This PR does not fully close this issue but works towards that effort, by creating a new benchmark tool using googlebench while making sure all functionality from current microbenchmark system is migrated to this new tool.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@ option(SLEEF_BUILD_QUAD "libsleefquad will be built." OFF)
 option(SLEEF_BUILD_GNUABI_LIBS "libsleefgnuabi will be built." ON)
 option(SLEEF_BUILD_SCALAR_LIB "libsleefscalar will be built." OFF)
 option(SLEEF_BUILD_TESTS "Tests will be built." ON)
+option(SLEEF_BUILD_BENCH "Bench will be built." OFF)
 option(SLEEF_BUILD_INLINE_HEADERS "Build header for inlining whole SLEEF functions" OFF)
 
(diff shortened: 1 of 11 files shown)


blapie:
The new runners are subject to rare random failures, so it is worth retriggering.
Go [here](https://github.com/shibatch/sleef/actions/runs/11272256486) and select retrigger failed jobs.

blapie:
> Go [here](https://github.com/shibatch/sleef/actions/runs/11272256486) and select retrigger failed jobs.

My bad you need specific permissions to do that. Doing it now.

blapie:
I will merge that unless there are some more comments.

This will make it easier to test and maintain, not to mention the extra benchmarking features. Looking forward to seeing new figures.

@shibatch Are we ok to go on with this solution and remove the pre-existing benchmarking suite?

PR #591, opened 2024-10-12
This patch updates helpers for ppc64el, s390x and purec_scalar.

shibatch:
It adds explicit typecasting so that C++ compilers can compile the inline headers generated from these helpers.

# Checklist

* This patch addresses the problem where the inline headers for ppc64el, s390x and purec_scalar cannot be compiled with C++ compilers.

# What changes did you make?

This patch updates helpers for ppc64el, s390x and purec_scalar.
It adds explicit typecasting so that C++ compilers can compile the inline headers generated from these helpers.

### DIFF
--- a/src/arch/helperpower_128.h
+++ b/src/arch/helperpower_128.h
@@ -103,16 +103,16 @@ typedef vquad vargquad;
 #define vset__s64(...) ((v__i64) {__VA_ARGS__})
 #define vset__u64(...) ((v__u64) {__VA_ARGS__})
 
-#define vsetall__vi(v)  vset__vi(v, v)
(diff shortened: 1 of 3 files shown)


blapie:
Hello! Aren't there intrinsics for power and s390x that could handle these operations in a vector fashion directly?

shibatch:
They were not available when I wrote those helpers.
They may be there now though.
I tried everything I could at the time.
Unlike Intel and ARM, reference manuals were not well maintained at that time, so I remember having a lot of trouble.

blapie:
Hm, ok. That's what I thought. It would be good to find someone that can look into these in the future.

Feel free to merge, so at least we can use inline headers in C++.

PR #592, opened 2024-10-16
This patch fixes two problems

shibatch:
# Checklist

* Fix bugs
  * On aarch32, cbrt(4) was returned instead of M_PI/2 when INFINITY is given to atanf.
  * Due to a bug in gcc-13, sign of tanf(+-0) was wrong with PUREC_SCALAR.

# What changes did you make?

* Changed a constant in sleefsimdsp.c to fix the value returned by xatanf when INFINITY is given on aarch32.
* Add a workaround that fixes the wrong sign of tanf with PUREC_SCALAR when compiled with gcc-13.
* Add testing for the first problem. The second problem is detected by the existing tester.

These changes are required to implement https://github.com/shibatch/sleef/issues/574, because these problems are also detected by the new tester.

### DIFF
--- a/src/libm/sleefsimdsp.c
+++ b/src/libm/sleefsimdsp.c
@@ -853,8 +853,8 @@ EXPORT CONST VECTOR_CC vfloat xtanf(vfloat d) {
   if (LIKELY(vtestallones_i_vo32(vlt_vo_vf_vf(vabs_vf_vf(d), vcast_vf_f(TRIGRANGEMAX2f*0.5f))))) {
     q = vrint_vi2_vf(vmul_vf_vf_vf(d, vcast_vf_f((float)(2 * M_1_PI))));
     u = vcast_vf_vi2(q);
-#if (defined(ENABLE_PUREC_SCALAR) || defined(ENABLE_PURECFMA_SCALAR)) && !defined(__clang__) && __GNUC__ == 13
-    if (q == 0) u = 0;
+#if (defined(ENABLE_PUREC_SCALAR) || defined(ENABLE_PURECFMA_SCALAR) || defined(ENABLE_VXE) || defined(ENABLE_VXENOFMA) || defined(ENABLE_VXE2) || defined(ENABLE_VXE2NOFMA)) && !defined(__clang__) && __GNUC__ == 13
+    u = vsel_vf_vo_vf_vf(veq_vo_vi2_vi2(q, vcast_vi2_i(0)), vcast_vf_f(0), u);
 #endif
     x = vmla_vf_vf_vf_vf(u, vcast_vf_f(-PI_A2f*0.5f), x);
     x = vmla_vf_vf_vf_vf(u, vcast_vf_f(-PI_B2f*0.5f), x);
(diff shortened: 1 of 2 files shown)


src/libm/sleefsimdsp.c
```diff
@@ -908,6 +911,9 @@ EXPORT CONST VECTOR_CC vfloat xtanf(vfloat d) {
 
   q = vrint_vi2_vf(vmul_vf_vf_vf(d, vcast_vf_f((float)(2 * M_1_PI))));
   u = vcast_vf_vi2(q);
+#if (defined(ENABLE_PUREC_SCALAR) || defined(ENABLE_PURECFMA_SCALAR) || defined(ENABLE_VXE) || defined(ENABLE_VXENOFMA) || defined(ENABLE_VXE2) || defined(ENABLE_VXE2NOFMA)) && !defined(__clang__) && __GNUC__ == 13
```
blapie, line 914: Is it just a problem with gcc 13?
If so we should seek to understand what is going on, because that might have repercussions elsewhere that we haven't noticed.

src/libm/sleefsimdsp.c
```diff
@@ -1770,7 +1776,7 @@ EXPORT CONST VECTOR_CC vfloat xatanf(vfloat d) {
   t = vreinterpret_vf_vm(vxor_vm_vm_vm(vand_vm_vo32_vm(veq_vo_vi2_vi2(vand_vi2_vi2_vi2(q, vcast_vi2_i(2)), vcast_vi2_i(2)), vreinterpret_vm_vf(vcast_vf_f(-0.0f))), vreinterpret_vm_vf(t)));
 
 #if defined(ENABLE_NEON32) || defined(ENABLE_NEON32VFPV4)
-  t = vsel_vf_vo_vf_vf(visinf_vo_vf(d), vmulsign_vf_vf_vf(vcast_vf_f(1.5874010519681994747517056f), d), t);
+  t = vsel_vf_vo_vf_vf(visinf_vo_vf(d), vmulsign_vf_vf_vf(vcast_vf_f(1.570796326794896557998982), d), t);
```
blapie, line 1779: Just a comment that in general I would recommend using hexfloat representation for constants, that would spare a conversion from double to float, or even from the literal to floating point.

Not asking to change now but that is smth we fix library-wide. That will also avoid carrying some many decimals, which is prone to errors.

Defining constants via macros could also prevent this type of errors.

src/libm/sleefsimdsp.c
blapie: If you wish to merge I would recommend creating an issue, so we don't forget about it.

blapie (commented):
Looks ok to me, just some general comments. Nothing that needs addressing urgently.

src/libm/sleefsimdsp.c
shibatch: There seems no problem with gcc-12 and gcc-14.
Of course it is better to understand what is going on, but that takes quite a bit of time.
I will create an issue for this matter.

src/libm/sleefsimdsp.c
shibatch: I agree that hexadecimal is better, but I was wondering which is better this time because of the MSVC case.
I will use hexadecimal next time.

Issue #579, opened 2024-09-20
Risc V and gcc-13


atupone:
See
https://bugs.gentoo.org/939400

It seems that sleef (3.6.1), on Risc-V, check for SIMD instruction that belongs to the v0.11.x draft
but then during build uses instructions in the v1.0.

gcc-13 only support the v0.11.x draft

I suggest checking instruction set belonging to the v1.0 (e.g. vfloat64m1x4_t)

Unfortunately I don't have a risc-v hardware to test

shibatch:
As for the RISC-V architecture, gcc-13 is not supported, see README.
I get so many errors when building with gcc-13. It is unlikely that gcc-13 will ever be supported.
I tried with gcc-14.2.0 and it passed the tests.

orlitzky:
Apparently it does build OK and pass tests once RVV is manually disabled -- we discovered this on the Gentoo bug.

I think all of the build failures are due to RVV being misdetected. With gcc-13, both RVV-1.0 and RVV-2.0 are detected but neither are implemented. This leads to build failures because sleef actually needs v1.0 to be fully implemented. The mis-detection with gcc-13 is due to the CMake tests that look for some types present in the v0.11.x draft of RVV. These are present in gcc-13 (so the test passes), but the v1.0 types aren't (so the build fails).

As @atupone suggests this can probably be fixed by checking for some types that are present in v1.0 and v2.0 only. I would be happy to test. I already have gcc-14 installed and it would be easy to install gcc-13 as well.

blapie:
@luhenry @GlassOfWhiskey @ericlove I thought I should draw your attention on this issue. Can you think of a way to make feature detection a bit more robust for RVV and avoid the mis-detection mentioned above for gcc 13+?

I will see what we can do regarding testing with gcc 13/14 in CI.

blapie:
@orlitzky Thanks for reporting the issue, feel free to suggest a fix. We will be able to test it locally with the appropriate compilers.

ericlove:
> @luhenry @GlassOfWhiskey @ericlove I thought I should draw your attention on this issue. Can you think of a way to make feature detection a bit more robust for RVV and avoid the mis-detection mentioned above for gcc 13+?

The standard way of testing for RVV v1.0 ISA compatibility would be to compare the `__riscv_v` macro against `1000000`, but in this case I think we should also compare the intrinsics API version against `12000` to ensure the tuple types are defined.

blapie:
[...]

PR #602, opened 2024-11-02
Configure.cmake: improve RVV1 check

orlitzky:
To test for RVV1 support, we currently run a small program that uses some vector intrinsics from the v0.11.x draft of the RVV standard. The library itself however needs newer intrinsics than the ones that we test for. This can lead to a problem because GCC 13 supports the intrinsics that we test for, but not (all of) the ones we intend to use. In short, the build can fail with GCC 13: https://github.com/shibatch/sleef/issues/579

This commit changes the detection mechanism. Instead of compiling a program that makes use of the intrinsics, we now just check the two values, `__riscv_v` and `__riscv_v_intrinsic`, to ensure that we have version 1.0 of RVV and version 0.12 of the intrinsics. According to the intrinsics spec, https://github.com/riscv-non-isa/rvv-intrinsic-doc, these are the stable versions.


ericlove (approved):
Looks great! Thanks for making this fix.

Issue #600, opened 2024-10-31
Exp and pow return infinity too early


MattyMuir:
**Describe the bug**
The `exp` function (for SIMD and pure C) returns infinity for all values past a certain threshold, but this threshold is slightly lower than it should be. Looking at the code for double precision, infinity is returned for all `x > 709.78271114955742909217217426`. However, there are values past this limit whose exponential can still be represented with a finite double. All standard libraries I've tested use the following, slightly larger, threshold instead `x > 709.782712893384` (`0x1.62e42fefa39efp+9` in binary). This doesn't appear to be an issue for the single precision implementation.

This same threshold is also used in the `pow` function, so the bug is present there as well.

**Command lines, logs and environment**
All the build steps, environment and configuration are the same as in my other issue, #570, though they shouldn't matter in this case.

**To Reproduce**
```C++
int main()
{
	double x = 709.782712;
	std::cout << "libc:  " << exp(x) << '\n';
	std::cout << "Sleef: " << Sleef_exp_u10(x) << '\n';
}
```
This produces the following output on my machine
```
libc:  1.79769e+308
Sleef: inf
```

Fixing this issue might be as simple as just changing the constant in the implementation, provided the series is still accurate past this point.

blapie:
Hello @MattyMuir, thanks for reporting this!

We will try to adjust the threshold so it is a little closer to general expectation. I also think that constants should be represented using hexfloat, they currently use too many digits, which I don't think is necessary.

Please consider the 2 following notes just FYI

Note 1: It's an interesting question, as I don't think there is any standard for this. Because we are working with approximation in finite precision, I think it is understandable (within standard) to overflow a little early or late and not behave precisely like the infinite precision exp, that being said if it is not a huge difference in performance it is probably preferable to fix it. In this case there is no performance difference because we already test against a threshold, it's the threshold that's a little off.

Note 2: Other math libraries like Arm Optimized routines have a more elaborate approach, it evacuates the early overflow to a separate branch where it is fixed, see https://github.com/ARM-software/optimized-routines/blob/master/math/aarch64/sv_exp_1u5.c.

blapie:
Just edited my notes above, as the first one was mistaken.

MattyMuir:
Just had a look at some of the single precision functions. Looks like similar bugs are present in `log1pf`, `sinhf`, and `coshf`. I think the tester needs to be updated to catch these, especially if overflow is mandated by the IEEE standard.

blapie:
[...]

PR #606, opened 2024-12-05
Update overflow bound in log1p(f)

### LINKED ISSUE #473: FP32 `log1pf_u10` and FP64 `log1p_u10` are infinitely inaccurate for large inputs
On x86 targets and possibly others, the `log1pf_u10` and `log1p_u10` functions are infinitely inaccurate for large input values.  Currently

- `log1pf_u10 (x > 1e38) = inf`, and
- `log1p_u10 (x > 1e307) = inf`

No input value except `x = inf` should result in an `inf` result for `log1p`.

Relevant source:

https://github.com/shibatch/sleef/blob/85440a5e87dae36ca1b891de14bc83b441ae7c43/src/libm/sleefsp.c#L1761

https://github.com/shibatch/sleef/blob/85440a5e87dae36ca1b891de14bc83b441ae7c43/src/libm/sleefdp.c#L2179

blapie:
# Checklist

* Fix a bug and add test to catch such issue in the future

# What changes did you make?

This PR consists in fix overflow behaviour in log1p(f) scalar and SIMD.

Fixes #473
Relates to #600 (where it is also reported)

# Is there anything you would like reviewers to focus on?

Please on focus on method chosen to fix approximation between old oflow threshold and max float/double.
Interval was thoroughly tested for log1pf and log1p, but testing is always as good as it gets to double precision.

### DIFF
--- a/src/common/misc.h
+++ b/src/common/misc.h
@@ -143,9 +143,16 @@
 
 #define R_LN2f 1.442695040888963407359924681001892137426645954152985934135449406931f
 
-// Overflow bound for exp and pow
+// Overflow bounds
 
+// - exp(x) overflows for x over (also used in pow)
 #define LOG_DBL_MAX 0x1.62e42fefa39efp+9 /* 709.782712893384 */
+
+// Other bounds
+
+// - log1p(f)(x) approximation holds up to x equals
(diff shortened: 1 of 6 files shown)


blapie:
Note that performance drops only slightly due to introduction of branch, INFINITY case is now delegated to log.

Issue #438, opened 2022-01-14
problem about simd xexpfrexp


longbiao7498:
There seems to be something wrong with xexpfrexp. when input is 2.1445984715023983335e-308,the output of libm's frexp and scalr expfrexp of sleef is -1022 while simd expfrexp of sleef is -969. And the vector implementation of xexpfrexp is inconsistent with its scalar implementation. @shibatch
Below is my fix:
```C
EXPORT CONST VECTOR_CC vint xexpfrexp(vdouble x) {
  vopmask o = vlt_vo_vd_vd(vabs_vd_vd(x), vcast_vd_d(DBL_MIN));
  x = vsel_vd_vo_vd_vd(o, vmul_vd_vd_vd(x, vcast_vd_d((double)(UINT64_C(1) << 63))), x);

  vint ret = vcastu_vi_vi2(vreinterpret_vi2_vd(x));
  ret = vsub_vi_vi_vi(vand_vi_vi_vi(vsrl_vi_vi_i(ret, 20), vcast_vi_i(0x7ff)), vcast_vi_i(0x3fe));
  o = vcast_vo32_vo64(o);
  ret = vsel_vi_vo_vi_vi(o, vadd_vi_vi_vi(vcast_vi_i(-63), ret), ret);

  vopmask o1 = vor_vo_vo_vo(vor_vo_vo_vo(veq_vo_vd_vd(x, vcast_vd_d(0)), visnan_vo_vd(x)), visinf_vo_vd(x));
  o1 = vcast_vo32_vo64(o1);
  ret = vsel_vi_vo_vi_vi(o1, vcast_vi_i(0), ret);
  return ret;
}
```

PR #627, opened 2025-02-11
fix  xexpfrexp

longbiao7498:
This commit fixes a bug in the xexpfrexp function: before calling vsel_vi_vo_vi_vi(), the xexpfrexp function did not use vcast_vo32_vo64() to convert vopmask64 to vopmask32. This omission caused incorrect results in the AdvSIMD implementation of xexpfrexp, making it unable to correctly handle input vectors that contain NaN and 0.0.


longbiao7498:
same with #438

longbiao7498:
This commit addresses the same issue described in issue #438

shibatch:
Thank you for your PR.
I will merge your patch after confirming that it passes all the tests.

PR #631, opened 2025-02-14
Add QNX SDP 8.0, 7.1 Support

eleir9268:
Used in pytorch mobile build. See instructions at https://github.com/qnx-ports/build-files/tree/pytorch_qnx_main .

It doesn't seem like I can get the tester executables to build for cross comp (because libmpfr does not come with QNX) so I did not test the built library (apart from verifying pytorch itself).


shibatch:
As for the tester, I recently added tester4 which can be used as a replacement of tester.
Please try adding -DSLEEF_ENABLE_TESTER4=True option to cmake.

shibatch:
We are sorry, but QNX support will be removed prior to the release of SLEEF 4.0.
As a general rule, we will have a policy of removing code that cannot be tested within our project, so this is a move in line with that policy.
Thank you for your understanding.

https://github.com/shibatch/sleef/discussions/673

PR #670, opened 2025-03-22
Use ISO 60559 extended floating point support instead of non-standard __float128

### LINKED ISSUE #479: unknown type name ‘__float128’ on loongarch64
image

image

wszqkzqk:
* Use ISO 60559's _Float to avoid platform compatibility risks
* https://en.cppreference.com/w/cpp/types/floating-point
* https://gcc.gnu.org/onlinedocs/gcc/Floating-Types.html
* Closes #479


shibatch:
It seems that gcc supports it, but I am concerned about the status of support in other compilers.

wszqkzqk:
> It seems that gcc supports it, but I am concerned about the status of support in other compilers.

`_Float128` is in ISO/IEC TS 18661-3:2015, while `__float128` is gcc extension instead. If we want better support in other compilers, we may avoid gcc extension `__float128`.

> ISO/IEC TS 18661-3:2015 defines C support for additional floating types _Floatn and _Floatnx, and GCC supports these type names; the set of types supported depends on the target architecture. Constants with these types use suffixes fn or Fn and fnx or Fnx. These type names can be used together with _Complex to declare complex types.

See also cppreference:

| Types<br>Defined in header<br><stdfloat> | Literal suffix | Predefined macro         | C language type | bits of storage | bits of precision | bits of exponent | max exponent |
|-----------------------------------------------|----------------|--------------------------|-----------------|-----------------|-------------------|------------------|--------------|
| float16_t                                     | f16 or F16     | __STDCPP_FLOAT16_T__     | _Float16        | 16              | 11                | 5                | 15           |
| float32_t                                     | f32 or F32     | __STDCPP_FLOAT32_T__     | _Float32        | 32              | 24                | 8                | 127          |
| float64_t                                     | f64 or F64     | __STDCPP_FLOAT64_T__     | _Float64        | 64              | 53                | 11               | 1023         |
| float128_t                                    | f128 or F128   | __STDCPP_FLOAT128_T__    | _Float128       | 128             | 113               | 15               | 16383        |
| bfloat16_t                                    | bf16 or BF16   | __STDCPP_BFLOAT16_T__    | (N/A)           | 16              | 8                 | 8                | 127          |

shibatch:
Yes, but ideals and reality are different, and as a practical matter, if __float128 is better supported, we might want to move on to _Float128 later.

shibatch:
You mentioned a C++ reference, but the important thing here is the support status in the C language.
I did a little searching, but the support status in C is unclear and needs to be checked properly.

wszqkzqk:
> Yes, but ideals and reality are different, and as a practical matter, if __float128 is better supported, we might want to move on to _Float128 later.

But it seems that the support of gcc extension`__float128` is more confusing, even in gcc itself. The suffix ‘q’ or ‘Q’ also may not be supported.

See also: [GCC Bugzilla: LoongArch: Q Suffix for __float128 Literals Not Supported](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119408)

GCC developer [Richard Biener](mailto:rguenth@gcc.gnu.org) told me that:

> Q is an extension, 1.11111f128 should work.  The extension is enabled only for some platforms.

wszqkzqk:
> You mentioned a C++ reference, but the important thing here is the support status in the C language. I did a little searching, but the support status in C is unclear and needs to be checked properly.

It's just a website to view the standard of both C and C++. The `C language type` is also in the table above.

shibatch:
Do you know anything about how well it is supported in Clang?

wszqkzqk:
The problem seems to have become more complicated 😵‍💫😵‍💫😵‍💫: https://github.com/llvm/llvm-project/issues/80195

Clang seems to support `__float128` only on x86_64, but it even hasn't support C23's `_Float128` nor `f128` suffix yet.

shibatch:
The best thing to do is to support both, and check __Float128 first, then check __float128 support if detection fails.
However, this is quite tedious.

shibatch:
Please create an issue. I will do this later.
This is quite complicated.

wszqkzqk:
> Please create an issue. I will do this later. This is quite complicated.

Please see #671

shibatch:
Good. Thanks.

PR #697, opened 2025-08-31
Add internal parallel-for framework to DFT

shibatch:
This PR adds an experimental framework for parallel execution of jobs.
This is to enable the use of all DFT features without OpenMP, as OpenMP tends to cause problems.
It is still experimental, and it's performance is not very stable.


dg0yt:
Nice. (Just struggling with vcpkg's sleef-rubberband-openmp.)

dg0yt:
Understood. 
Being able to build DFT again with stock AppleClang without extra OpenMP libs might be beneficial regardless of ultimate performance. The rubberband port in vcpkg switches between different sleef and fftw3, subject to platform. fftw3 is GPL-licensed.

